PR #12, opened 2019-03-07
allow publishing of artifacts

roastario:


build.gradle
```diff
@@ -51,8 +74,64 @@ allprojects {
...
+                pom {
+                    description = subproject.description
+                }
+            }
+        }
+    }
+}
+
 
-    jar {
```
roger3cev: We need this to exclude log4j2.xml from the JAR

modules/money/build.gradle
```diff
@@ -1,3 +1,16 @@
...
+    dependencies {
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+        classpath "net.corda.plugins:cordapp:$corda_gradle_plugins_version"
+        classpath "net.corda.plugins:quasar-utils:$corda_gradle_plugins_version"
+        classpath "org.jetbrains.kotlin:kotlin-noarg:$kotlin_version"
+    }
+}
+
+apply plugin: 'kotlin-jpa'
+apply plugin: 'net.corda.plugins.quasar-utils'
```
roger3cev: dont think we need quasar utils here. we can actually remove kotlin JPA here too, I think.

PR #13, opened 2019-03-11
Updated the code to create session only if the recipient is not the node initiating the flow

ajithathayahar:
1. IssueToken -> Checked and created the session only if the node running the flow is different from the owner. Please check whether I have created owningParty correctly in case if anonymous is set to true
2. UpdateEvolvableToken -> Filtered the updateSubscribers list in case if the node running the flow is part of the updateSubscribers list. This will happen, if we have self-issued token.
3.  MoveToken -> When a token is transferred to a party, I felt that new party should be added to the distribution list hence added a subflow AddPartyToDistributionList. If this is not needed, please let me know.

### DIFF
--- a/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/IssueToken.kt
+++ b/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/IssueToken.kt
@@ -72,18 +72,21 @@ object IssueToken {
         @Suspendable
         override fun call(): SignedTransaction {
             // This is the identity which will be used to issue tokens.
-            // We also need a session for the other side.
             val me: Party = ourIdentity
-            val ownerSession = initiateFlow(owner)
-
-            // Notify the recipient that we'll be issuing them a tokens and advise them of anything they must do, e.g.
-            // generate a confidential identity for the issuer or sign up for updates for evolvable tokens.
-            ownerSession.send(TokenIssuanceNotification(anonymous = anonymous))
-
+            var ownerSession : FlowSession? = null
             // This is the recipient of the tokens identity.
-            val owningParty: AbstractParty = if (anonymous) {
-                subFlow(RequestConfidentialIdentity.Initiator(ownerSession)).party.anonymise()
-            } else owner
+            val owningParty: AbstractParty
+    
+            if(me != owner) {
+                // We also need a session for the other side.
+                ownerSession = initiateFlow(owner)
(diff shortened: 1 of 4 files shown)


workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/IssueToken.kt
```diff
@@ -72,18 +72,21 @@ object IssueToken {
...
-            // Notify the recipient that we'll be issuing them a tokens and advise them of anything they must do, e.g.
-            // generate a confidential identity for the issuer or sign up for updates for evolvable tokens.
-            ownerSession.send(TokenIssuanceNotification(anonymous = anonymous))
-
+            var ownerSession : FlowSession? = null
             // This is the recipient of the tokens identity.
-            val owningParty: AbstractParty = if (anonymous) {
-                subFlow(RequestConfidentialIdentity.Initiator(ownerSession)).party.anonymise()
-            } else owner
+            val owningParty: AbstractParty
```
roger3cev: maybe this can be a conditional assignment? `val owningParty = if (x) { y() } else { z() }`

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/MoveToken.kt
```diff
@@ -40,7 +41,11 @@ object MoveToken {
             val owningParty: AbstractParty = if (anonymous) {
                 subFlow(RequestConfidentialIdentity.Initiator(ownerSession)).party.anonymise()
             } else owner
-
+    
+            if (ownedToken is TokenPointer<*>) {
```
roger3cev: I'm not sure how this will work. I set up the distribution list so that only the creator of the evolvable token type has the distribution list, as they are the only ones which can update the list. Here we are setting up a distribution list on the party which is moving the token (which references the evolvable token type) to another node. So they will have a list but there is no code path in the token SDK currently which would result in this non-issuer party forwarding on the evolvable token updates.... This is where data distribution groups come in. I realise that for now, the functionality here is quite limited until we get data distribution groups. The solution is for all `Party`s to manually sign up with the maintainer of the evolvable token type using the `RequestAdditionToDistributionList` flow.

ajithathayahar: Shall i remove this AddPartyToDistributionList?  I misunderstood the concept and to enable all parties receiving the updates when an update happens on the evolvable token, I added this.

PR #17, opened 2019-03-18
start working on an inmemory token selector

roastario:


roger3cev:
Looks as though it's going in the right direction! Can we catch-up on this, perhaps with @rick-r3 ? Cheers

PR #18, opened 2019-03-19
Changed single to singleOrNull when checking that issuer is sole signer of an issuance

shobande-femi:
When `issueCommand.signers` contains more than one element, calling
`single` will throw an `IllegalArgumentException("List has more than one element.")`
which translates to `TransactionVerificationException.ContractRejection`
being thrown.

Without this fix, [this](https://github.com/corda/token-sdk/blob/2a8d9199b07baea554d49b8fe2937b927b7b4ce0/contract/src/test/kotlin/com/r3/corda/sdk/token/contracts/FungibleTokenTests.kt#L81) assertion fails
<br>
_I hereby certify that my contribution is in accordance with the Developer Certificate of Origin (https://developercertificate.org/)._


roger3cev:
Ah.. thanks for the contribution. You are indeed correct that we get the wrong message if more than one key or zero keys are supplied. Cheers!

PR #21, opened 2019-03-25
repackage under com.r3.corda.lib namespace

roastario:


build.gradle
```diff
@@ -37,7 +37,7 @@ plugins {
 
 allprojects {
     version "1.0-SNAPSHOT"
-    group "com.r3.tokens-sdk"
+    group "com.r3.corda.lib"
```
josecoll: this feels very abstract - I'd like to see **tokens** kept in the namespace.

PR #25, opened 2019-04-01
Add in memory token selection

roastario:


rick-r3:
Will take a look today

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/selection/VaultWatcherService.kt
```diff
@@ -0,0 +1,148 @@
...
+
+    constructor(appServiceHub: AppServiceHub) : this(getObservableFromAppServiceHub(appServiceHub))
+
+    companion object {
+        val LOG = contextLogger()
+
+        private fun getObservableFromAppServiceHub(appServiceHub: AppServiceHub): VaultObserver {
+            val pageSize = 1000
+            var currentPage = DEFAULT_PAGE_NUM;
+            var (existingStates, observable) = appServiceHub.vaultService.trackBy(FungibleToken::class.java, PageSpecification(pageNumber = currentPage, pageSize = pageSize))
```
rick-r3: You might want to replace this with a JPA query directly for `PersistentFungibleToken` to initialise at start up to make things a lot faster.  If that table can hold the fields needed to bucket these, plus the amount, that would make the init sequence much faster.  Loading all of the whole states via the vault will be very expensive.

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/selection/VaultWatcherService.kt
```diff
@@ -0,0 +1,148 @@
...
+import java.security.PublicKey
+import java.time.Duration
+import java.util.concurrent.*
+
+val unlocker: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
+
+@CordaService
+class VaultWatcherService(vaultObserver: VaultObserver? = null) {
+
+    val cache: ConcurrentMap<PublicKey, ConcurrentMap<Class<*>, ConcurrentMap<String, ConcurrentMap<StateAndRef<FungibleToken<TokenType>>, Boolean>>>> = ConcurrentHashMap()
```
rick-r3: Try using some types/classes!

Why not flatten the key to be the one you build in `processToken` ?  All the nested maps will be inefficient from a memory perspective.  Don't you just want `ConcurrentMap<BucketIdentifier, Bucket>`.

`processToken` becomes `fun generateBucketId(token: FungibleToken<*>): BucketIdentifier`

And for now `Bucket` just encapsulates the `ConcurrentMap<StateAndRef<FungibleToken<TokenType>>, Boolean>` - which I think should be changed anyway.... see other comments.

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/selection/VaultWatcherService.kt
```diff
@@ -0,0 +1,148 @@
...
+            owner: PublicKey,
+            amountRequested: Amount<IssuedTokenType<T>>,
+            predicate: ((StateAndRef<FungibleToken<TokenType>>) -> Boolean) = { true },
+            allowSubSelect: Boolean = false,
+            autoUnlockDelay: Duration = Duration.ofMinutes(5)
+    ): List<StateAndRef<FungibleToken<T>>> {
+        val set = getTokenSet(owner, amountRequested.token.tokenType.tokenClass, amountRequested.token.tokenType.tokenIdentifier)
+        val lockedTokens = mutableListOf<StateAndRef<FungibleToken<TokenType>>>()
+        var amountLocked: Amount<IssuedTokenType<T>> = amountRequested.copy(quantity = 0)
+        for (tokenStateAndRef in set.keys) {
```
rick-r3: Not totally sure iterating on a `ConcurrentHashMap` key Set is necessarily a good idea.  What ordering does it have?  No idea how well it performs.  Can you get a concurrent modification exception?

Anyway, all this would be functions on `Bucket` class.

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/selection/VaultWatcherService.kt
```diff
@@ -0,0 +1,148 @@
...
+        val (owner, type, typeId) = processToken(token)
+        val tokensForTypeInfo = getTokenSet(owner, type, typeId)
+        tokensForTypeInfo.replace(stateAndRef, true, false)
+    }
+
+    inline fun <T : TokenType> selectTokens(
+            owner: PublicKey,
+            amountRequested: Amount<IssuedTokenType<T>>,
+            predicate: ((StateAndRef<FungibleToken<TokenType>>) -> Boolean) = { true },
+            allowSubSelect: Boolean = false,
```
rick-r3: `allowShortfall` ?  Sub select made me think of SQL stuff.

PR #31, opened 2019-04-09
Some minor cleanup and additions

roger3cev:
* Renamed Hack contract to "DummyContract"
* Added helpers for all major currencies and added a helper for XRP.


modules/money/src/main/kotlin/com/r3/corda/sdk/token/money/DummyContract.kt
```diff
@@ -7,6 +7,6 @@ import net.corda.core.transactions.LedgerTransaction
  * This is included so that the CorDapp scanner auto-magically includes this JAR in the attachment store. It will remain
  * until CorDapp dependencies are properly handled in Corda 5.0.
  */
-class ThisIsAnnoying : Contract {
+class DummyContract : Contract {
```
kasiastreich: Very good :joy:

roger3cev: 😋

manosbatsis: A proper change, but i'll miss the original classname from the output of deployNodes...

PR #34, opened 2019-04-11
Alternative map based selection

roastario:


workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/selection/VaultWatcherService.kt
```diff
@@ -0,0 +1,143 @@
...
+import rx.Observable
+import java.lang.RuntimeException
+import java.security.PublicKey
+import java.time.Duration
+import java.util.concurrent.*
+
+val unlocker: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
+
+@CordaService
+class VaultWatcherService(vaultObserver: VaultObserver? = null) {
```
roger3cev: does this need to sub-class `SingletonSerializeAsToken` ?

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/selection/VaultWatcherService.kt
```diff
@@ -0,0 +1,143 @@
...
+                existingStates = appServiceHub.vaultService.queryBy(FungibleToken::class.java, PageSpecification(pageNumber = ++currentPage, pageSize = pageSize))
+            }
+            return VaultObserver(listOfThings, observable)
+        }
+
+        fun processToken(token: FungibleToken<*>): TokenIndex {
+            val owner = token.holder.owningKey
+            val type = token.amount.token.tokenType.tokenClass
+            val typeId = token.amount.token.tokenType.tokenIdentifier
+            return TokenIndex(owner, type, typeId)
```
roger3cev: How come you are indexing by owning key? This could potentially be different for each token, right? So you'd end up with a key in the cache for _each_ token? I get why you'd want to do it by x500 name though. Presumably you just need to index by token type (class and ID) to get a bucket per token type. Then if we want to index by, say... notary, issuer, or whatever, then we can add other index types?

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/selection/VaultWatcherService.kt
```diff
@@ -0,0 +1,143 @@
...
+                        break
+                    }
+                }
+            }
+        }
+        if (!allowSubSelect && amountLocked < amountRequested) {
+            throw InsufficientBalanceException("Could not find enough tokens to satisfy token request")
+        }
+
+        unlocker.schedule({
```
roger3cev: so if tokens dont get used they are auto unlocked after 5 minutes. Seems like a long time. I wonder if there is a clever way to lock tokens to a flow ID, then watch for when that flow ending successfully or with an error. If it's successful then we remove from set. If it's an error, then we unlock?

PR #37, opened 2019-04-11
Make addPartyToDistributionList an utility function

kasiastreich:


workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/utilities/AddPartyToDistributionList.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.r3.corda.sdk.token.workflow.schemas.DistributionRecord
+import net.corda.core.contracts.UniqueIdentifier
+import net.corda.core.identity.Party
+import net.corda.core.node.ServiceHub
+
+/**
+ * Utility function to persist a new entity pertaining to a distribution record.
+ * TODO: Add some error handling.
+ */
+fun ServiceHub.addPartyToDistributionList(party: Party, linearId: UniqueIdentifier) {
```
roger3cev: Does it have to be an extension function? Just thinking of Java users :)

kasiastreich: Good point. It doesn't will change in next PR ;)

PR #39, opened 2019-04-15
Misc changes

kasiastreich:
* Change access to query utilities.
* Allow to call MoveToken flow to confidential identity on the same node.
* Add changeOwner to generateMove utility function.


roger3cev:
Merging #39

PR #40, opened 2019-04-15
Fix

kasiastreich:
I definitely need more coffee

### DIFF
--- a/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/selection/TokenSelection.kt
+++ b/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/selection/TokenSelection.kt
@@ -188,10 +188,10 @@ class TokenSelection(val services: ServiceHub, private val maxRetries: Int = 8, 
             // TODO This approach is clearly taken from core. Revisit it.
             services.keyManagementService.freshKeyAndCert(nodeIdentity, revocationEnabled = false).party.anonymise()
         } else {
-            // Check that this identity belongs to the node that called generateMove
+            // Check that this identity belongs to the node that called generateMove.
             val ownerId = services.identityService.wellKnownPartyFromAnonymous(changeOwner)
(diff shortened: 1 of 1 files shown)


workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/selection/TokenSelection.kt
```diff
@@ -188,10 +188,10 @@ class TokenSelection(val services: ServiceHub, private val maxRetries: Int = 8,
             // TODO This approach is clearly taken from core. Revisit it.
             services.keyManagementService.freshKeyAndCert(nodeIdentity, revocationEnabled = false).party.anonymise()
         } else {
-            // Check that this identity belongs to the node that called generateMove
+            // Check that this identity belongs to the node that called generateMove.
```
roger3cev: right so... if no key is provided then a new confidential key is created for the node doing token selection. Otherwise, if there is a key, then we check that the key belongs to the node doing token selection (so we don't send change to someone else!).

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/selection/TokenSelection.kt
```diff
@@ -188,10 +188,10 @@ class TokenSelection(val services: ServiceHub, private val maxRetries: Int = 8,
...
             val ownerId = services.identityService.wellKnownPartyFromAnonymous(changeOwner)
-            check(ownerId != null && !services.myInfo.isLegalIdentity(ownerId)) {
-                "Owner of the change: $changeOwner is the identity that belongs to the node."
+            check(ownerId != null && services.myInfo.isLegalIdentity(ownerId)) {
+                "Owner of the change: $changeOwner is not the identity that belongs to the node."
```
roger3cev: Surely this should be "$changeOwner IS an identity that belongs to the node."

PR #41, opened 2019-04-16
Bump version to 5.0-SNAPSHOT to experiment with shell serialisers.

roger3cev:
* Bump version to 5.0-SNAPSHOT to experiment with shell serialisers.
* Add a dependency on node-api to the money module so we can write some custom deserialisers. This is hopefully only temporary.
* Cleaned up dupe dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 buildscript {
     ext {
         corda_release_group = 'net.corda'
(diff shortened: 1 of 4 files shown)


roger3cev:
Closing this as it turns out the change wont be available for ages, so no point in trying it.

PR #47, opened 2019-04-23
Split the Issue, Move and Redeem flows into the general and an extra custom flow

IgorWolkov:
One failed test `DiamondWithTokenScenarioTests`, potential cause: anonymized parties handled in a wrong way.

### DIFF
--- a/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/IssueToken.kt
+++ b/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/IssueToken.kt
@@ -12,19 +12,15 @@ import com.r3.corda.sdk.token.contracts.utilities.withNotary
 import com.r3.corda.sdk.token.workflow.utilities.addPartyToDistributionList
 import net.corda.core.contracts.Amount
 import net.corda.core.contracts.TransactionState
-import net.corda.core.flows.FinalityFlow
-import net.corda.core.flows.FlowLogic
-import net.corda.core.flows.FlowSession
-import net.corda.core.flows.InitiatedBy
-import net.corda.core.flows.InitiatingFlow
-import net.corda.core.flows.ReceiveFinalityFlow
-import net.corda.core.flows.StartableByRPC
+import net.corda.core.contracts.requireThat
+import net.corda.core.flows.*
 import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.Party
 import net.corda.core.node.StatesToRecord
 import net.corda.core.transactions.SignedTransaction
 import net.corda.core.transactions.TransactionBuilder
 import net.corda.core.utilities.ProgressTracker
+import java.security.PublicKey
 
 /**
  * This flow takes a bunch of parameters and is used to issue a token or an amount of some token on the ledger to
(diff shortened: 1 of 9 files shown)


IgorWolkov:
@illoyd, @kasiastreich, @roger3cev could you please share your thoughts about the approach. If it's the right way, then I'll merge new master into the branch and prepare cleaner pr.
Thanks.

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/IssueToken.kt
```diff
@@ -12,19 +12,15 @@ import com.r3.corda.sdk.token.contracts.utilities.withNotary
...
 import net.corda.core.contracts.TransactionState
-import net.corda.core.flows.FinalityFlow
-import net.corda.core.flows.FlowLogic
-import net.corda.core.flows.FlowSession
-import net.corda.core.flows.InitiatedBy
-import net.corda.core.flows.InitiatingFlow
-import net.corda.core.flows.ReceiveFinalityFlow
-import net.corda.core.flows.StartableByRPC
+import net.corda.core.contracts.requireThat
+import net.corda.core.flows.*
```
illoyd: Try not to use * imports

workflow/src/test/kotlin/com/r3/corda/sdk/token/workflow/FlowHelpers.kt
```diff
@@ -26,7 +26,7 @@ fun <T : TokenType> StartedMockNode.issueTokens(
         issueTo: StartedMockNode,
         notary: StartedMockNode,
         amount: Amount<T>? = null,
-        anonymous: Boolean = true
+        anonymous: Boolean = false
```
illoyd: Try not to change default values; other tests rely on them to be set a certain way and may impact test states and performance if changed.

workflow/src/test/kotlin/com/r3/corda/sdk/token/workflow/TokenFlowWithTradeConditionsTests.kt
```diff
@@ -0,0 +1,70 @@
+package com.r3.corda.sdk.token.workflow
```
illoyd: Good test driven design.

workflow/src/test/kotlin/com/r3/corda/sdk/token/workflow/flows/IssueTokenWitTradeConditions.kt
```diff
@@ -0,0 +1,56 @@
+package com.r3.corda.sdk.token.workflow.flows
```
illoyd: Be careful with file names. This one has a typo.

workflow/src/test/kotlin/com/r3/corda/sdk/token/workflow/statesAndContracts/TradeConditions.kt
```diff
@@ -0,0 +1,90 @@
...
+import net.corda.core.schemas.QueryableState
+import net.corda.core.serialization.CordaSerializable
+import net.corda.core.transactions.LedgerTransaction
+import java.util.*
+import javax.persistence.Column
+import javax.persistence.Entity
+import javax.persistence.Table
+
+@BelongsToContract(TradeConditionsContract::class)
+data class TradeConditions(val owner: AbstractParty,
```
illoyd: Instead of `TradeConditions` being a simple string, how about converting it to an encumbrance? This way you can model how an actual state constraint would work in Corda. (Of course, be aware of the caveats of not mixing an encumbrance with a reference state; this will work fine for tokens with a fixed (inline) definition but not for tokens using an EvolvableToken token type.)

PR #48, opened 2019-04-23
MoveToken should update subscriptions

kasiastreich:
Simplified solution for updating the subscription list held by maintainers. There are many problems with this approach and we should move to data distribution lists, but this should do for now.

### DIFF
--- a/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/MoveToken.kt
+++ b/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/MoveToken.kt
@@ -29,7 +29,8 @@ object MoveToken {
 
     @InitiatingFlow
     @StartableByRPC
-    abstract class Initiator<T : TokenType>(
+    abstract class Initiator<T: TokenType>(
+            val token: T,
             val holder: AbstractParty,
             val session: FlowSession? = null
     ) : FlowLogic<SignedTransaction>() {
@@ -66,8 +67,8 @@ object MoveToken {
             val sessions = if (ourIdentity == holderParty) emptyList() else listOf(holderSession)
             val finalTx = subFlow(FinalityFlow(transaction = stx, sessions = sessions))
             // If it's TokenPointer, then update the distribution lists for token maintainers
-            if(ownedToken is TokenPointer<*>) {
-                subFlow(UpdateDistributionList.Initiator(ownedToken, ourIdentity, holderParty))
+            if(token is TokenPointer<*>) {
+                subFlow(UpdateDistributionList.Initiator(token, ourIdentity, holderParty))
             }
             return finalTx
         }
@@ -76,7 +77,7 @@ object MoveToken {
     @InitiatedBy(Initiator::class)
     class Responder(val otherSession: FlowSession) : FlowLogic<Unit>() {
         @Suspendable
-        override fun call(): Unit {
+        override fun call() {
             // Resolve the issuance transaction.
             if (!serviceHub.myInfo.isLegalIdentity(otherSession.counterparty)) {
                 subFlow(ReceiveFinalityFlow(otherSideSession = otherSession, statesToRecord = StatesToRecord.ONLY_RELEVANT))
@@ -89,7 +90,8 @@ class MoveTokenNonFungible<T: TokenType>(
         val ownedToken: T,
         holder: AbstractParty,
         session: FlowSession? = null
-): MoveToken.Initiator<T>(holder, session) {
+): MoveToken.Initiator<T>(ownedToken, holder, session) {
+    @Suspendable
     override fun generateMove(): Pair<TransactionBuilder, List<PublicKey>> {
         return generateMoveNonFungible(serviceHub.vaultService, ownedToken, holder)
     }
@@ -99,7 +101,8 @@ open class MoveTokenFungible<T: TokenType>(
         val amount: Amount<T>,
         holder: AbstractParty,
         session: FlowSession? = null
-): MoveToken.Initiator<T>(holder, session) {
+): MoveToken.Initiator<T>(amount.token, holder, session) {
+    @Suspendable
     override fun generateMove(): Pair<TransactionBuilder, List<PublicKey>> {
         val tokenSelection = TokenSelection(serviceHub)
         return tokenSelection.generateMove(TransactionBuilder(), amount, holder)
--- a/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/MoveToken.kt
+++ b/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/MoveToken.kt
@@ -5,8 +5,6 @@ import com.r3.corda.sdk.token.contracts.types.TokenPointer
 import com.r3.corda.sdk.token.contracts.types.TokenType
 import com.r3.corda.sdk.token.workflow.selection.TokenSelection
 import com.r3.corda.sdk.token.workflow.selection.generateMoveNonFungible
-import com.r3.corda.sdk.token.workflow.utilities.ownedTokenAmountCriteria
-import com.r3.corda.sdk.token.workflow.utilities.tokenAmountWithIssuerCriteria
 import net.corda.core.contracts.Amount
 import net.corda.core.flows.FinalityFlow
 import net.corda.core.flows.FlowLogic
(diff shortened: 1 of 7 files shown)


workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/UpdateDistributionList.kt
```diff
@@ -0,0 +1,93 @@
...
+    ) : FlowLogic<Unit>() {
+        @Suspendable
+        override fun call() {
+            val evolvableToken = tokenPointer.pointer.resolve(serviceHub).state.data
+            val distributionListUpdate = DistributionListUpdate(oldParty, newParty, evolvableToken.linearId)
+            val maintainers = evolvableToken.maintainers
+            val maintainersSessions = maintainers.map { initiateFlow(it) }
+            // Collect signatures from old and new parties
+            val updateBytes = distributionListUpdate.serialize()
+            val ourSig = serviceHub.keyManagementService.sign(updateBytes.bytes, oldParty.owningKey)
```
roger3cev: do we need the signing part? Seems like a bit overkill :) What attack vector is the singing preventing?

roger3cev: *signing

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/UpdateDistributionList.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    @CordaSerializable
+    data class DistributionListUpdate(val oldParty: Party, val newParty: Party, val linearId: UniqueIdentifier)
+
+    // TODO It's ineffective if we are heavily using confidential identities, because we contact maintainers every time we do move
+    // tokens with token pointers.
+    @InitiatingFlow
+    class Initiator<T : EvolvableTokenType>(
+            val tokenPointer: TokenPointer<T>,
+            val oldParty: Party,
```
roger3cev: Don't need this if we are not doing signing

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/UpdateDistributionList.kt
```diff
@@ -0,0 +1,93 @@
...
+                }
+                // Check that the request comes from that party.
+                check(update.oldParty == otherSession.counterparty) {
+                    "Got distribution list update request from a counterparty: ${otherSession.counterparty} " +
+                            "that isn't a signer of request: ${update.oldParty}."
+                }
+                update
+            }
+            // Check that newParty is well known party.
+            serviceHub.identityService.wellKnownPartyFromAnonymous(distListUpdate.newParty)
```
roger3cev: you can use `requireWellKnownPartyFromAnonymous`

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/UpdateDistributionList.kt
```diff
@@ -0,0 +1,93 @@
...
+                serviceHub.addPartyToDistributionList(distListUpdate.newParty, distListUpdate.linearId)
+            }
+        }
+
+        @Suspendable
+        private fun getDistributionRecord(linearId: UniqueIdentifier, party: Party): List<DistributionRecord> {
+            return serviceHub.withEntityManager {
+                val query: CriteriaQuery<DistributionRecord> = criteriaBuilder.createQuery(DistributionRecord::class.java)
+                query.apply {
+                    val root = from(DistributionRecord::class.java)
```
roger3cev: is there a utility function for this? If not maybe we can create one

roger3cev: See `getDistributionList` in the workflow module. Maybe we can add another method. Cheers!

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/UpdateEvolvableTokenResponder.kt
```diff
@@ -25,6 +25,6 @@ class UpdateEvolvableTokenResponder(val otherSession: FlowSession) : FlowLogic<S
         }
 
         // Resolve the creation transaction.
-        return subFlow(ReceiveFinalityFlow(otherSideSession = otherSession, statesToRecord = StatesToRecord.ONLY_RELEVANT))
+        return subFlow(ReceiveFinalityFlow(otherSideSession = otherSession, statesToRecord = StatesToRecord.ALL_VISIBLE))
```
roger3cev: This change is because, before, the other participants were not recording the updates? I think it was only relevant before because there was only one maintainer but now that's not the case. Good spot.

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/UpdateDistributionList.kt
kasiastreich: You are right, initially I was wondering if we should remove old owners (that's why signed data structure), but it doesn't make much sense if it's fungible token.

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/UpdateDistributionList.kt
kasiastreich: The method you mentioned returns a list of parties, I want just a single record that matches pair (linearId, party) - it's more optimal than loading a list and searching for party.

kasiastreich: Ah, I just read first comment, yes, I will add it to utilities.

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/UpdateEvolvableTokenResponder.kt
kasiastreich: Yeah, it was failing some tests.

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/UpdateDistributionList.kt
roger3cev: We should remove this before V1 is released

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/UpdateDistributionList.kt
```diff
@@ -0,0 +1,63 @@
...
+import net.corda.core.flows.InitiatedBy
+import net.corda.core.flows.InitiatingFlow
+import net.corda.core.identity.Party
+import net.corda.core.serialization.CordaSerializable
+import net.corda.core.utilities.unwrap
+
+object UpdateDistributionList {
+
+    @CordaSerializable
+    data class DistributionListUpdate(val oldParty: Party, val newParty: Party, val linearId: UniqueIdentifier)
```
roger3cev: likewise remove old party here before v1 is released.

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/MoveToken.kt
```diff
@@ -63,15 +65,19 @@ object MoveToken {
...
         }
     }
 
-    // TODO Don't really need it anymore as it calls only finality flow
     @InitiatedBy(Initiator::class)
     class Responder(val otherSession: FlowSession) : FlowLogic<Unit>() {
         @Suspendable
-        override fun call(): Unit {
+        override fun call() {
             // Resolve the issuance transaction.
```
roger3cev: move as opposed to issuance.

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/flows/MoveToken.kt
```diff
@@ -94,7 +101,8 @@ open class MoveTokenFungible<T: TokenType>(
         val amount: Amount<T>,
```
roger3cev: we should all catch up and figure out what the optimal patterns for structuring issuance, move, redeem flows.

workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/utilities/AddPartyToDistributionList.kt
```diff
@@ -8,6 +8,7 @@ import net.corda.core.node.ServiceHub
 /**
  * Utility function to persist a new entity pertaining to a distribution record.
  * TODO: Add some error handling.
+ * TODO: Don't duplicate pairs of linearId and party.
```
roger3cev: This probably doesn't matter for now - before this ever becomes an issue we will transition to DDGs.

PR #51, opened 2019-05-07
make rowsToAmount in QueryUtilities public to allow for re-use

philippkueng:
We'd like to use this function when writing our own `VaultService` extension functions.
Is there anything speaking against making it public?


roger3cev:
Oh and thanks for the contribution!  👍

roastario:
triggered merge of master

roastario:
@philippkueng sorry, but could you close and re-submit this PR? 

It seems to have triggered some sort of error condition with our TeamCity integration, where it thinks it has built the branch, but never does..

PR #54, opened 2019-05-13
Change AbstractToken<*>.holderString to explicit, dynamic String.

jplodge-pro:
Fix to TSK-19 **IFF** a bug. 
Changed`TokenUtilities.kt`, to convert `val AbstractToken<*>.holderString` to an explicit `String` and not a `() -> String`, as appears to be expected by all the usages.


jplodge-pro:
Couldn't have my name on the list without an actual contribution!

PR #55, opened 2019-05-13
FungibleTokenContract.verifyMove now allows multiple move commands

jplodge-pro:
Fixes TSK-18 **IFF** it is a bug.
1. Edited `FungibleTokenContract.verifyMove` to retrieve the group Token Type from the `first()` `MoveCommand` instead of the `single()` such command.
2. Attempted a Move Token contract test to confirm results.

**NB** I made this PR because we encountered this problem during testing and it seemed a simple fix. However, my test case **DOES NOT** reproduce the behaviour we encountered exactly and I don't know how to make it do so.  
Our function composes a single transaction from a set of unrelated bilateral move token "sub-transactions". This manifests, at `verifyMove` time, as a `moveCommands` list that has >1 entry.  Thus, the invocation of `single()` was causing contract verification failure.


roger3cev:
From what I gather, this _is_ a bug. Thanks for the fix!

PR #57, opened 2019-05-13
make rowsToAmount in QueryUtilities public to allow for re-use (2nd attempt)

philippkueng:
re-open due to @roastario's request https://github.com/corda/token-sdk/pull/51

### DIFF
--- a/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/utilities/QueryUtilities.kt
+++ b/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/utilities/QueryUtilities.kt
@@ -119,7 +119,7 @@ fun sumTokenCriteria(): QueryCriteria {
 }
 
 // Abstracts away the nasty 'otherResults' part of the vault query API.
-private fun <T : TokenType> rowsToAmount(token: T, rows: Vault.Page<FungibleToken<T>>): Amount<T> {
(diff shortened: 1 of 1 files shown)


roger3cev:
@philippkueng  This is _supposed_ to be a private function. What do you need to expose it for? Or rather... What other public functions are missing? Cheers

philippkueng:
@roger3cev that was my question as part of PR https://github.com/corda/token-sdk/pull/51 (the same diff) - we're needing it because we'd like to create our own account specific queries and would like to re-use your base queries, enhance the criteria list and then use `rowsToAmount` as if we'd be just using your queries as part of the token-sdk.

roger3cev:
OK great. So what specific queries do you need? We can add them to the project?

roger3cev:
The thing is... I'm reluctant to expose loads of detail in the SDK because I need the freedom to change implementation in the future. Exposing this will lead to being quite limited on what we can do in the future. I'd far rather add public APIs which handle your needs.

philippkueng:
I understand your concern and in some respects given that token-sdk isn't 1.0 stable yet, we currently consider all of it changeable anyway. Will attempt to commit some of the queries back to the token-sdk and potentially account-sdk in the future (@nimmaj will have more details on) - and if you feel like you have to change it, then please do that. In that case we can still decide if we'd like to adapt or then copy your previous version into our code.

roger3cev:
@philippkueng Thanks for understanding! Will approve this for now so it unblocks you. Cheers

PR #65, opened 2019-05-23
Change softLockType to UNLOCKED_ONLY

kasiastreich:
Multiple calls to addMove in the same flow (with the same lockId) should select next unlocked tokens.


roger3cev:
This PR breaks one of the unit tests. I'll investigate but I'll probably end up closing this as I don't think it's a good idea.

roger3cev:
actually it's fine... it's a behaviour change that we will need to get used to... Need to release soft locks in unit tests if you are doing multiple selections in the same test.

PR #67, opened 2019-05-23
Issue 66 fix.

roger3cev:
Don't add duplicate records to the distribution list when issuing tokens with evolvable token types.

### DIFF
--- a/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/utilities/FlowUtilities.kt
+++ b/workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/utilities/FlowUtilities.kt
@@ -22,12 +22,12 @@ import java.security.PublicKey
  * TODO: Add some error handling.
  */
 @Suspendable
-fun FlowLogic<*>.addPartyToDistributionList(services: ServiceHub, party: Party, linearId: UniqueIdentifier) {
+fun FlowLogic<*>.addPartyToDistributionList(party: Party, linearId: UniqueIdentifier) {
     // Create an persist a new entity.
-    val hasRecord = hasDistributionRecord(services, linearId, party)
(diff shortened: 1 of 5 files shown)


workflow/src/main/kotlin/com/r3/corda/sdk/token/workflow/utilities/FlowUtilities.kt
```diff
@@ -22,10 +22,16 @@ import java.security.PublicKey
  * TODO: Add some error handling.
  */
 @Suspendable
-fun addPartyToDistributionList(services: ServiceHub, party: Party, linearId: UniqueIdentifier) {
+fun FlowLogic<*>.addPartyToDistributionList(services: ServiceHub, party: Party, linearId: UniqueIdentifier) {
```
kasiastreich: When you extend flow logic then you don't need to add services as parameter

workflow/src/test/kotlin/com/r3/corda/sdk/token/workflow/TokenFlowTests.kt
```diff
@@ -155,7 +155,7 @@ class TokenFlowTests : MockNetworkTest(numberOfNodes = 4) {
         val issueTokenA = I.issueFungibleTokens(A, 50 of housePointer).getOrThrow()
         A.watchForTransaction(issueTokenA.id).toCompletableFuture().getOrThrow()
         // Issue to node B.
-        val issueTokenB = I.issueFungibleTokens(A, 50 of housePointer).getOrThrow()
+        val issueTokenB = I.issueFungibleTokens(B, 50 of housePointer).getOrThrow()
         A.watchForTransaction(issueTokenB.id).toCompletableFuture().getOrThrow()
```
kasiastreich: B should watch for transaction after this change

roger3cev:
Addressed comments @kasiastreich cheers!

PR #75, opened 2019-06-12
Add integration test with simple DvP

kasiastreich:


workflow/src/test/kotlin/com/r3/corda/sdk/token/workflow/TokenDriverTest.kt
```diff
@@ -0,0 +1,241 @@
...
+            // Move house to Node B using conf identities in exchange for cash using DvP flow.
+            // TODO use conf identities
+            val dvpTx = nodeA.rpc.startFlowDynamic(
+                    DvPFlow::class.java,
+                    house,
+                    nodeBParty
+            ).returnValue.getOrThrow()
+
+            // Wait for both nodes to record the transaction.
+            nodeA.rpc.watchForTransaction(dvpTx).getOrThrow(5.seconds)
```
roger3cev: do we really want timeouts here?

roger3cev:
Will have to force merge this as the CI doesn't work....

PR #76, opened 2019-06-13
Fixed not copying across old linear ID when calling `withNewHolder`.

roger3cev:
Thanks to @johnbuckle for finding this.

Also fixed the evolvable token type tests now we bumped the version to V5.


roger3cev:
Thanks! Force merging as CI is busted.

PR #77, opened 2019-06-13
Publishing changes 

roger3cev:
- Changed module names from "contract", "workflow" to "contracts" and "workflows", so inline with other CorDapps. 
- Now don't publish an empty "modules" JAR. 
- Changed package namespace to "com.r3.corda.lib". 
- Changed artefact IDs to be prefixed with "tokens-".


roastario:
LGTM!

roger3cev:
CI is broken so force merging.

PR #80, opened 2019-06-17
CORDA-3015: Compile contracts against corda-core-deterministic instead of corda-core.

chrisr3:
Build the `:contracts` module against `corda-core-deterministic` to ensure that it doesn't reference any non-deterministic functionality. This should help prepare it to be sandboxed by the DJVM.


roastario:
This is a Test from Jenkins

roastario:
build

chrisr3:
It's _green_ :green_heart:. But you're going to want to examine what I've done with the `infix` functions before merging it.

chrisr3:
AFAIK the `publish` blocks are for the sake of the `net.corda.plugins.publish-utils` plugin. Which the Tokens SDK doesn't use... :thinking:

You could equally ask whether me removing the `humphrey` blocks will affect publication :grin:.

PR #89, opened 2019-06-24
constrain TokenType JAR

roastario:


tudor-malene:
I was expecting a new field added to the `AbstractToken` state. Which would have been propagated from input to output states, and then the jar from the transaction attachments checked against it.

Unless I'm missing something, the code in this pr just reads the jar from the classloader, and then checks it against all the other states that read from the same classloader, which means there's no security check on it.

roastario:
> I was expecting a new field added to the `AbstractToken` state. Which would have been propagated from input to output states, and then the jar from the transaction attachments checked against it.
> 
> Unless I'm missing something, the code in this pr just reads the jar from the classloader, and then checks it against all the other states that read from the same classloader, which means there's no security check on it.

@tudor-malene We do enforce the check that the jar hash never changes via the `com.r3.corda.lib.tokens.contracts.AbstractTokenContract#verifyAllTokensUseSameTypeJar`
which enforces the fact that all inputs and outputs *must* use the same hash (and jar)

tudor-malene:
> > I was expecting a new field added to the `AbstractToken` state. Which would have been propagated from input to output states, and then the jar from the transaction attachments checked against it.
> > Unless I'm missing something, the code in this pr just reads the jar from the classloader, and then checks it against all the other states that read from the same classloader, which means there's no security check on it.
> 
> @tudor-malene We do enforce the check that the jar hash never changes via the `com.r3.corda.lib.tokens.contracts.AbstractTokenContract#verifyAllTokensUseSameTypeJar`
> which enforces the fact that all inputs and outputs _must_ use the same hash (and jar)

But of course it doesn't change if you read the jar from the same classloader for both the input and the output states.

roastario:
> > > I was expecting a new field added to the `AbstractToken` state. Which would have been propagated from input to output states, and then the jar from the transaction attachments checked against it.
> > > Unless I'm missing something, the code in this pr just reads the jar from the classloader, and then checks it against all the other states that read from the same classloader, which means there's no security check on it.
> > 
> > 
> > @tudor-malene We do enforce the check that the jar hash never changes via the `com.r3.corda.lib.tokens.contracts.AbstractTokenContract#verifyAllTokensUseSameTypeJar`
> > which enforces the fact that all inputs and outputs _must_ use the same hash (and jar)
> 
> But of course it doesn't change if you read the jar from the same classloader for both the input and the output states.


OK, maybe I am missing something but 

Issuer issues token with hash = [ABCFD]

from then on all tokens of this type must be also [ABCFD] no?

The detail is that the tokens contract will "group" all tokens of a given type before applying this rule - so it should be possible to enforce that the token jar must always be the same as when it was issued.

tudor-malene:
If you don’t encode the hash in the state, you have to load the previous transaction of all input states and read the hash from that classloader

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

roastario:
> If you don’t encode the hash in the state, you have to load the previous transaction of all input states and read the hash from that classloader

Ah, I realise what you mean now. 

Yes sorry, I have changed it to a `val` instead of a `fun` in the current version of the code, I forgot to repush.

contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/states/FungibleToken.kt
```diff
@@ -31,9 +33,10 @@ import net.corda.core.schemas.QueryableState
  * @param T the [TokenType] this [FungibleToken] state is in respect of.
  */
 @BelongsToContract(FungibleTokenContract::class)
-open class FungibleToken<T : TokenType>(
+class FungibleToken<T : TokenType>(
```
kasiastreich: NonFungibleToken is still open with one final field is it on purpose?

deterministic.gradle
```diff
@@ -1,14 +1,14 @@
...
-private final void deterministic(Configuration configuration) {
-    if (configuration.state == Configuration.State.UNRESOLVED) {
-        // Ensure that this module uses the deterministic Corda artifacts.
-        configuration.resolutionStrategy.dependencySubstitution {
-            substitute module("$corda_release_group:corda-serialization:$corda_release_version") with module("$corda_release_group:corda-serialization-deterministic:$corda_release_version")
-            substitute module("$corda_release_group:corda-core:$corda_release_version") with module("$corda_release_group:corda-core-deterministic:$corda_release_version")
-        }
-    }
-}
+//configurations {
```
kasiastreich: cleanup

modules/test-contracts/src/main/kotlin/com/r3/corda/lib/tokens/testing/states/Rubles.kt
```diff
@@ -0,0 +1,28 @@
+package com.r3.corda.lib.tokens.testing.states
+
+import com.r3.corda.lib.tokens.contracts.types.TokenType
+
+
+open class Ruble : TokenType {
```
kasiastreich: oh man, wrong currency, polski złoty

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/flows/issue/IssueTokensUtilities.kt
```diff
@@ -45,4 +46,28 @@ fun addIssueTokens(transactionBuilder: TransactionBuilder, vararg outputs: Abstr
 @Suspendable
 fun addIssueTokens(transactionBuilder: TransactionBuilder, output: AbstractToken<*>): TransactionBuilder {
     return addIssueTokens(transactionBuilder, listOf(output))
+}
+
+// Utilities for ensuring that the correct JAR which implements TokenType is added to the transaction.
+
+fun addTokenTypeJar(tokens: List<AbstractToken<*>>, transactionBuilder: TransactionBuilder) {
```
kasiastreich: Should we move it to internal package?

workflows/src/test/kotlin/com/r3/corda/lib/tokens/workflows/JITMockNetworkTests.kt
```diff
@@ -85,4 +88,11 @@ abstract class JITMockNetworkTests(val names: List<CordaX500Name> = emptyList())
      * Smart helper for [assertNotHasTransaction] that passes in the test suite's network.
      */
     protected fun assertNotHasTransaction(tx: SignedTransaction, vararg nodes: StartedMockNode) = assertNotHasTransaction(tx, network, *nodes)
+
+}
+
+fun convertToStartedMockNode(createPartyNode: TestStartedNode): StartedMockNode {
```
kasiastreich: What is it?

contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/states/FungibleToken.kt
roger3cev: We can open it up later if necessary.

modules/test-contracts/src/main/kotlin/com/r3/corda/lib/tokens/testing/states/Rubles.kt
roger3cev: :laugh

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/flows/issue/IssueTokensUtilities.kt
roger3cev: Others that create their own flows might need to use this

workflows/src/test/kotlin/com/r3/corda/lib/tokens/workflows/JITMockNetworkTests.kt
roger3cev: @roastario  ?

roger3cev:
Looks like tests are passing and we have approval from @tudor-malene !

PR #107, opened 2019-07-05
Remove all generic TokenTypes where possible

roastario:


modules/contracts-for-testing/src/main/kotlin/com/r3/corda/lib/tokens/testing/states/Rubles.kt
```diff
@@ -11,8 +11,8 @@ import net.corda.core.identity.AbstractParty
 import net.corda.core.transactions.LedgerTransaction
 
 @BelongsToContract(FungibleTokenContract::class)
-open class RubleToken(override val amount: Amount<IssuedTokenType<RUB>>,
-                      override val holder: AbstractParty) : FungibleToken<RUB>(amount, holder)
+open class RubleToken(amount: Amount<IssuedTokenType>,
```
roger3cev: How come you are sub-classing `FungibleToken` here? Shouldn't we be sub-classing `TokenType` ?

contracts/src/test/kotlin/com/r3/corda/lib/tokens/contracts/TokenPointerTests.kt
```diff
@@ -50,15 +49,15 @@ class TokenPointerTests : ContractTestCommon() {
         }
         // Check we can resolve the pointer inside the ledger transaction.
         val ledgerTransaction = testTransaction.toLedgerTransaction(aliceServices)
-        val fungibleToken = ledgerTransaction.singleOutput<FungibleToken<TokenPointer<TestEvolvableTokenType>>>()
-        assertEquals(fungibleToken.tokenType.pointer.resolve(ledgerTransaction), outputStateAndRef)
+        val fungibleToken = ledgerTransaction.singleOutput<FungibleToken>()
+        assertEquals(((fungibleToken.tokenType) as TokenPointer<TestEvolvableTokenType>).pointer.resolve(ledgerTransaction), outputStateAndRef)
```
roger3cev: can we just remove the type parameter for token pointer?

roger3cev: Can just set it to `EvolvableTokenType`. We lose some type saftey but makes things a bit easier ?

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/internal/flows/distribution/DistributionUtilities.kt
```diff
@@ -62,21 +62,21 @@ fun ServiceHub.addPartyToDistributionList(party: Party, linearId: UniqueIdentifi
 }
 
 @Suspendable
-fun FlowLogic<*>.addToDistributionList(tokens: List<AbstractToken<TokenPointer<*>>>) {
-    tokens.forEach { token ->
-        val tokenType = token.tokenType
+fun FlowLogic<*>.addToDistributionList(tokens: List<AbstractToken>) {
+    tokens.filter { it.tokenType as? TokenPointer<*> != null }.forEach { token ->
```
roger3cev: assumption here is that all tokens coming this way have token pointers... we should probably refactor this code a bit as it's a little confusing...! Either way it's all going when we get DDGs.

modules/contracts-for-testing/src/main/kotlin/com/r3/corda/lib/tokens/testing/states/Rubles.kt
roastario: this is for a specific edge case in testing where I wanted to enforce that a `FungibleToken` cannot be moved to a `SubclassedFungibleToken` during a move

contracts/src/test/kotlin/com/r3/corda/lib/tokens/contracts/NonFungibleTokenTests.kt
```diff
@@ -21,7 +20,7 @@ class NonFungibleTokenTests : ContractTestCommon() {
     fun `issue non fungible token tests`() {
         transaction {
             // Start with only one output.
-            output(NonFungibleTokenContract.contractId, issuedToken heldBy ALICE.party)
+            output(NonFungibleTokenContract.contractId, issuedToken.heldBy(ALICE.party))
```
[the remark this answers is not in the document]
roastario: will revert, this was midway through generic hell.

contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/utilities/TransactionUtilities.kt
```diff
@@ -85,7 +85,7 @@ internal val NULL_SECURE_HASH = SecureHash.zeroHash
  */
 fun TokenType.getAttachmentIdForGenericParam(): SecureHash? {
     val computedValue = synchronized(attachmentCache) {
-        val startingPoint = if (this is IssuedTokenType<*>) {
+        val startingPoint = if (this is IssuedTokenType) {
```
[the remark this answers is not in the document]
roastario: yes, but this method is also called by tests which operate on raw TokenTypes

contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/states/FungibleToken.kt
```diff
@@ -33,21 +33,21 @@ import net.corda.core.schemas.QueryableState
  * @param T the [TokenType] this [FungibleToken] state is in respect of.
  */
 @BelongsToContract(FungibleTokenContract::class)
-open class FungibleToken<T : TokenType>(
-        override val amount: Amount<IssuedTokenType<T>>,
+open class FungibleToken(
+        override val amount: Amount<IssuedTokenType>,
         override val holder: AbstractParty,
         override val tokenTypeJarHash: SecureHash? = amount.token.tokenType.getAttachmentIdForGenericParam()
```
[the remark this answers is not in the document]
roastario: it needs to be a constructor value, otherwise it is "evaluated" when got. This means that within the verify sandbox, it would be operating on whatever classes were provided to the TX. 

That would allow people to bypass the pinning.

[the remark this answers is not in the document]
roastario: IE, the value must be persisted to the ledger rather than a computed value at runtime.

contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/commands/TokenCommand.kt
```diff
@@ -18,7 +18,7 @@ import net.corda.core.contracts.CommandData
  * @property token the group of [IssuedTokenType]s this command should be tied to.
  * @param T the [TokenType].
  */
-abstract class TokenCommand<T : TokenType>(open val token: IssuedTokenType<T>, internal val inputIndicies: List<Int> = listOf(), internal val outputIndicies: List<Int> = listOf()) : CommandData {
+abstract class TokenCommand(open val token: IssuedTokenType, internal val inputIndicies: List<Int> = listOf(), internal val outputIndicies: List<Int> = listOf()) : CommandData {
```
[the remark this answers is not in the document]
roastario: technically true, but that would involve a large amount of changes to the contracts - I would feel safer doing that _after_ the V1 release.

contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/states/NonFungibleToken.kt
```diff
@@ -32,25 +32,25 @@ import net.corda.core.schemas.QueryableState
  *
  * @property token the [IssuedTokenType] which this [NonFungibleToken] is in respect of.
  * @property holder the [AbstractParty] which holds the [IssuedTokenType].
- * @param T the [TokenType].
+ * @param TokenType the [TokenType].
```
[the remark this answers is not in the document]
roastario: done

contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/states/FungibleToken.kt
```diff
@@ -33,21 +33,21 @@ import net.corda.core.schemas.QueryableState
  * @param T the [TokenType] this [FungibleToken] state is in respect of.
```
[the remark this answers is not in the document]
roastario: done

contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/states/NonFungibleToken.kt
```diff
@@ -32,25 +32,25 @@ import net.corda.core.schemas.QueryableState
  *
  * @property token the [IssuedTokenType] which this [NonFungibleToken] is in respect of.
  * @property holder the [AbstractParty] which holds the [IssuedTokenType].
```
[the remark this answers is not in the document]
roastario: done

contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/types/IssuedTokenType.kt
```diff
@@ -22,7 +22,7 @@ import java.math.BigDecimal
  * @param T the [TokenType].
```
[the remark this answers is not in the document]
roastario: done

modules/contracts-for-testing/src/main/kotlin/com/r3/corda/lib/tokens/testing/states/Rubles.kt
```diff
@@ -47,8 +47,8 @@ data class Appartment(val id: String = "Foo") : TokenType(id, 0)
 
 
 @BelongsToContract(DodgeTokenContract::class)
-open class DodgeToken(override val amount: Amount<IssuedTokenType<RUB>>,
-                      override val holder: AbstractParty) : FungibleToken<RUB>(amount, holder)
+open class DodgeToken(amount: Amount<IssuedTokenType>,
```
kasiastreich: This doesn't seem right, that DodgeToken theoretically should hold Rubles, but there is no Ruble in type

roastario: There are no types any more! `DodgeToken` can now hold any TokenType.

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/internal/flows/distribution/DistributionUtilities.kt
kasiastreich: You can use filterIsInstance, then you don't need to do a cast later

modules/contracts-for-testing/src/main/kotlin/com/r3/corda/lib/tokens/testing/states/Rubles.kt
kasiastreich: yeah, I know, that's why i am wondering why there is a DodgeToken in a first place :D

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/internal/flows/distribution/DistributionUtilities.kt
roastario: Achually, you can't because we are not filtering on `it` but rather `it.tokenType` so Kotlin isn't smart enough to remember what `it.tokenType` is.

contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/commands/TokenCommand.kt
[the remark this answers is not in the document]
roger3cev: can we remove it after V1 though? Need to maintain compatibility

contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/utilities/TransactionUtilities.kt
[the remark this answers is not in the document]
roger3cev: ah ok thanks

modules/contracts-for-testing/src/main/kotlin/com/r3/corda/lib/tokens/testing/states/Rubles.kt
roastario: I needed a token that had a different contract to test the condition that tokens cannot move to a type with a different contract

contracts/src/test/kotlin/com/r3/corda/lib/tokens/contracts/TokenPointerTests.kt
roastario: done

roastario:
> This is looking good. The only issue I have is that now it's difficult to know what type of `TokenType` is contained within a token state... Is it a pointer or an inlined token type? You now have to attempt a cast which introduces a bit of complexity that the type param hid. It's not a huge deal though, we can create some utilities which help... @kasiastreich @roastario what you reckon?

What are the conditions where you could have a Token
1. You are creating a token, so you know exactly what the types are (castable)
2. You have queried from the vault, so you know what you are querying for (castable)
3. Any other condition - the generics dont help! at best they will throw a runtime exception.

roastario:
@roger3cev actually, I believe this PR has broken querying of Tokens by their pointer class.

https://github.com/corda/token-sdk/pull/107/files#diff-1f74ecdc8bc2e98736184c0247ad0f34R49 should work right?

roger3cev:
Will have a look this weekend/monday. I think other than this we are good to go unless anything else crops up in the meantime. Cheers

roger3cev:
I added a new branch: https://github.com/corda/token-sdk/pull/110

roger3cev:
I've reset the branch to https://github.com/corda/token-sdk/pull/107/commits/2775457b340229b546b2fed727e238a37316fe5c. Thanks for giving this a shot but looks like

PR #110, opened 2019-07-08
Rog revert token pointer changes

roger3cev:
Takes @roastario 's branch but resets back 2 commits and adds a new commit.


contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/states/NonFungibleToken.kt
```diff
@@ -32,25 +32,27 @@ import net.corda.core.schemas.QueryableState
  *
  * @property token the [IssuedTokenType] which this [NonFungibleToken] is in respect of.
  * @property holder the [AbstractParty] which holds the [IssuedTokenType].
- * @param T the [TokenType].
+ * @property linearId the [UniqueIdentifier] which will uniquely identify this Token.
+ * @property tokenTypeJarHash the [SecureHash] which will pin the jar that provides the [TokenType].
```
kasiastreich: Maybe mention that devs should use default and there is no need to provide it

contracts/src/test/kotlin/com/r3/corda/lib/tokens/contracts/ContractTestCommon.kt
```diff
@@ -68,12 +68,12 @@ abstract class ContractTestCommon {
 
 /**
  * Creates a [NonFungibleToken] from an [IssuedTokenType].
- * E.g. IssuedTokenType<TokenType> -> NonFungibleToken<TokenType>.
+ * E.g. IssuedTokenType<TokenType> -> NonFungibleToken.
  * This function must exist outside of the contracts module as creating a unique identifier is non deterministic.
  */
-infix fun <T : TokenType> IssuedTokenType<T>.heldBy(owner: AbstractParty): NonFungibleToken<T> = _heldBy(owner)
+infix fun IssuedTokenType.heldBy(owner: AbstractParty): NonFungibleToken = _heldBy(owner)
```
kasiastreich: You have 2 functions that do the same now, but only takes Party and other AbstractParty. Why? see: https://github.com/corda/token-sdk/pull/110/files#diff-69f95b21984486950778649bdf50c5b0R66

contracts/src/test/kotlin/com/r3/corda/lib/tokens/contracts/NonFungibleTokenTests.kt
```diff
@@ -143,7 +143,7 @@ class NonFungibleTokenTests : ContractTestCommon() {
                 val anotherIssuedTokenHeldByAlice = anotherIssuedToken heldBy ALICE.party
 
                 input(NonFungibleTokenContract.contractId, anotherIssuedTokenHeldByAlice)
-                output(NonFungibleTokenContract.contractId, anotherIssuedTokenHeldByAlice withNewHolder BOB.party)
+                output(NonFungibleTokenContract.contractId, anotherIssuedTokenHeldByAlice.withNewHolder(BOB.party))
```
kasiastreich: do we still have withNewHolder function? Or was it changed to withHolder? It's inconsistent

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenUtilities.kt
```diff
@@ -4,17 +4,16 @@ package com.r3.corda.lib.tokens.workflows.utilities
...
 import net.corda.core.identity.AbstractParty
 
 /**
  * Creates a [NonFungibleToken] from an [IssuedTokenType].
- * E.g. IssuedTokenType<TokenType> -> NonFungibleToken<TokenType>.
+ * E.g. IssuedTokenType<TokenType> -> NonFungibleToken.
  * This function must exist outside of the contracts module as creating a unique identifier is non deterministic.
  */
-infix fun <T : TokenType> IssuedTokenType<T>.heldBy(owner: AbstractParty): NonFungibleToken<T> = _heldBy(owner)
+infix fun IssuedTokenType.heldBy(owner: AbstractParty): NonFungibleToken = _heldBy(owner)
```
kasiastreich: Why do i see this function for a third time :joy:

roger3cev: Has to be here as we can't use `UniqueIdentifier` inside the `contracts` module.

contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/states/NonFungibleToken.kt
roger3cev: OK

contracts/src/test/kotlin/com/r3/corda/lib/tokens/contracts/ContractTestCommon.kt
roger3cev: Chris Rankin moved these. We need them here because they cannot be defined inside the `contracts` main module. They are not part of the public API so we can find a better solution at some point in the future.

contracts/src/test/kotlin/com/r3/corda/lib/tokens/contracts/NonFungibleTokenTests.kt
roger3cev: There is only one function `Abstract.withNewHolder`. There is also `heldBy` which is an infix function and used when creating `AbstratToken`s.

contracts/src/test/kotlin/com/r3/corda/lib/tokens/contracts/ContractTestCommon.kt
kasiastreich: and yet in TokenUtilties we have...
`infix fun IssuedTokenType.heldBy(party: Party): NonFungibleToken {
    return NonFungibleToken(this, party, UniqueIdentifier())
} `

PR #113, opened 2019-07-09
Cleanup docs after recent changes

kasiastreich:


docs/IWantTo.md
```diff
@@ -22,62 +22,27 @@ amounts of this token can have. E.g.
...
+    val myTokenType: TokenType = TokenType(tokenIdentifier = "TEST", fractionDigits = 2)
 ```
 
-
 The `tokenIdentifier` is used along with the `tokenClass` property (defined
 in `TokenType`) when serializing token types. Two properties are required,
 as with `FiatCurrency` and `DigitalCurrency`, there can be many different
 instances of one `tokenClass`, each with their own `tokenIdentifier`.
 
-The above defined token type, allows CorDapp developers to create multiple
```
roger3cev: we should keep this under another section which explains how to create your own token type sub-classes.

PR #124, opened 2019-07-22
Make evolvable token flows use ObserverAwareFinalityFlow

kasiastreich:


workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/flows/evolvable/CreateEvolvableTokensFlowHandler.kt
```diff
@@ -1,15 +1,17 @@
...
-import net.corda.core.flows.*
-import net.corda.core.node.StatesToRecord
+import net.corda.core.flows.FlowLogic
+import net.corda.core.flows.FlowSession
+import net.corda.core.flows.SignTransactionFlow
 import net.corda.core.transactions.SignedTransaction
 import net.corda.core.utilities.unwrap
 
-class CreateEvolvableTokensFlowHandler(val otherSession: FlowSession) : FlowLogic<SignedTransaction>() {
+class CreateEvolvableTokensFlowHandler(val otherSession: FlowSession) : FlowLogic<Unit>() {
```
roger3cev: Would we maybe want to return the signed transaction? Maybe the responder might need it for subsequent things in the parent flow?

workflows/src/test/kotlin/com/r3/corda/lib/tokens/workflows/factories/TestEvolvableTokenTypeFactory.kt
```diff
@@ -21,31 +21,31 @@ class TestEvolvableTokenTypeFactory(val testSuite: JITMockNetworkTests, val main
         )
     }
 
-    fun withOneMaintainerAndOneObserver(linearId: UniqueIdentifier = UniqueIdentifier(), maintainer: Party = testSuite.identity(maintainerA), observer: Party = testSuite.identity(observerC)): TestEvolvableTokenType {
+    fun withOneMaintainerAndOneParticipant(linearId: UniqueIdentifier = UniqueIdentifier(), maintainer: Party = testSuite.identity(maintainerA), observer: Party = testSuite.identity(observerC)): TestEvolvableTokenType {
```
roger3cev: How come the rename ? The function still has an `observer` parameter

workflows/src/test/kotlin/com/r3/corda/lib/tokens/workflows/CreateEvolvableTokenTests.kt
```diff
@@ -165,4 +166,65 @@ class CreateEvolvableTokenTests : JITMockNetworkTests() {
...
+        val createTx = alice.createEvolvableToken(token, notaryIdentity, listOf(denise.legalIdentity())).getOrThrow()
+        val createdToken = createTx.singleOutput<TestEvolvableTokenType>()
+
+        // Expect to have one create command with maintainer signature
+        val maintainerKeys = token.maintainers.map { it.owningKey }.toSet()
+        assertEquals(maintainerKeys, createTx.requiredSigningKeys, "Must be signed by all maintainers")
+
+        // Alice, Charlie, Denise should record the transaction
+        assertHasTransaction(createTx, alice, charlie, denise)
+        val aliceToken = alice.services.vaultService.queryBy<TestEvolvableTokenType>().states
```
roger3cev: I assume alice is the issuer here and the other two are the observer / participant?

workflows/src/test/kotlin/com/r3/corda/lib/tokens/workflows/CreateEvolvableTokenTests.kt
```diff
@@ -165,4 +166,65 @@ class CreateEvolvableTokenTests : JITMockNetworkTests() {
...
+        // Alice, Charlie, Denise should record the transaction
+        assertHasTransaction(createTx, alice, charlie)
+        val aliceToken = alice.services.vaultService.queryBy<TestEvolvableTokenType>().states
+        assertEquals(createdToken, aliceToken.single())
+        val charlieToken = charlie.services.vaultService.queryBy<TestEvolvableTokenType>().states
+        assertEquals(createdToken, charlieToken.single())
+    }
+
+    @Test
+    fun `with observer that is maintainer`() {
```
roger3cev: What's the difference between this test and the above one?

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/flows/evolvable/CreateEvolvableTokensFlowHandler.kt
kasiastreich: It's because of the return type of observer aware finality flow, I will check if I could work around that in the handler.

workflows/src/test/kotlin/com/r3/corda/lib/tokens/workflows/factories/TestEvolvableTokenTypeFactory.kt
kasiastreich: It's because it wasn't an observer in the oberverSessions sense. it is additional participant on evolvable token type. The definition of `TestEvolvableTokenType` uses observers only to add them to participants list.

workflows/src/test/kotlin/com/r3/corda/lib/tokens/workflows/CreateEvolvableTokenTests.kt
kasiastreich: It's confusing because I reused old states and tests to do that. Basically, observer is also maintainer of the state. In previous test observer is participant. They are different in terms of what I pass to the flow as an observers list.
I wanted to test that even if someone passes in observers that are already on the state (either as a maintainer or participant) the flows will still function correctly.
Maybe it needs some redesign though, I could use your input on how it should be used.

workflows/src/test/kotlin/com/r3/corda/lib/tokens/workflows/CreateEvolvableTokenTests.kt
kasiastreich: I added comments.

roger3cev: 👍

workflows/src/test/kotlin/com/r3/corda/lib/tokens/workflows/CreateEvolvableTokenTests.kt
roger3cev: ok thanks for the heads up - for now it's OK, I think.

PR #125, opened 2019-07-30
Added updated hierarchy diagram.

roger3cev:


roger3cev:
I think it's OK to leave it out as it would complicate the diagram quite a bit!

PR #129, opened 2019-08-23
ENT-4035 Rebase with new CI

willhr3:


workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/flows/rpc/MoveTokens.kt
```diff
@@ -134,8 +136,11 @@ class ConfidentialMoveFungibleTokens(
         val participants = partiesAndAmounts.map(PartyAndAmount<*>::party)
         val observerSessions = sessionsForParties(observers)
         val participantSessions = sessionsForParties(participants)
-        val confidentialHolder = changeHolder
-                ?: serviceHub.keyManagementService.freshKeyAndCert(ourIdentityAndCert, false).party.anonymise()
+        val confidentialHolder = changeHolder?:let {
```
roger3cev: Need to make sure it's properly documented here that if you are using confidential identities you need to generate the key and assign it to an ID BEFORE you call move tokens flow, otherwise a new random key will be generated but not associated to any ID.

roger3cev:
build

PR #130, opened 2019-08-29
External id indexing for in memory token selection

kasiastreich:
This PR should be reviewed in 2 commits.
First one moves LocalTokenSelection to separate module (although, we should have something like token-api to have it done cleanly so we have relevant interfaces (Selector and config) and common types live there.

Second one introduces 2 new indexing types for in memory token selection: just by token class and id (without owner) and by external id.
There is a terrible hack because bug in corda (there is not database transaction on observable thread, so we can't query publicKey -> external id mapping). Additionally, most of it should go after pk -> external id cache is introduced in Corda, if ever...
I introduced Holder sealed class to represent different types of token ownership indexing.


modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/SelectionUtilities.kt
```diff
@@ -0,0 +1,11 @@
+package com.r3.corda.lib.tokens.selection
```
roger3cev: maybe this should be `com.r3.corda.lib.tokens.workflows.selection` ?

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/selectors/Selector.kt
```diff
@@ -1,20 +1,18 @@
-package com.r3.corda.lib.tokens.workflows.internal.selection
```
roger3cev: DB selection is still in `workflows` shall we move it to this new module?

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/selectors/Selector.kt
```diff
@@ -53,7 +51,7 @@ interface Selector {
     @Suspendable
     fun generateMove(
             lockId: UUID = FlowLogic.currentTopLevel?.runId?.uuid ?: UUID.randomUUID(),
-            partiesAndAmounts: List<PartyAndAmount<TokenType>>,
+            partiesAndAmounts: List<Pair<AbstractParty, Amount<TokenType>>>,
```
roger3cev: How come we got rid of the `PartyAndAmount` type ?

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/config/InMemorySelectionConfig.kt
```diff
@@ -0,0 +1,60 @@
...
+            logger.info("Found in memory token selection configuration with values indexing strategy: $indexingType, cacheSize: $cacheSize")
+            return InMemorySelectionConfig(indexingType, cacheSize)
+        }
+    }
+
+    @Suspendable
+    override fun toSelector(services: ServiceHub): LocalTokenSelector {
+        return try {
+            val vaultObserver = services.cordaService(VaultWatcherService::class.java)
+            LocalTokenSelector(services, vaultObserver, state = null)// TODO allowShortFall and autoUnlockDelay, should it be config or per flow?
```
roger3cev: probably per flow but set to a sensible default?

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/config/InMemorySelectionConfig.kt
```diff
@@ -0,0 +1,60 @@
...
+data class InMemorySelectionConfig(val indexingStrategy: VaultWatcherService.IndexingType, val cacheSize: Int = CACHE_SIZE_DEFAULT) : StateSelectionConfig {
+    companion object {
+        val logger = LoggerFactory.getLogger("inMemoryConfigSelectionLogger")
+        fun parse(config: CordappConfig): InMemorySelectionConfig {
+            val cacheSize = config.getIntOrNull("stateSelection.inMemory.cacheSize")
+                    ?: CACHE_SIZE_DEFAULT
+            val indexingType = try {
+                VaultWatcherService.IndexingType.valueOf(config.get("stateSelection.inMemory.indexingStrategy").toString().toUpperCase())
+            } catch (e: CordappConfigException) {
+                VaultWatcherService.IndexingType.PUBLIC_KEY // Default is public key.
```
roger3cev: Let's make the default token type only

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/config/InMemorySelectionConfig.kt
```diff
@@ -0,0 +1,60 @@
...
+import co.paralleluniverse.fibers.Suspendable
+import com.r3.corda.lib.tokens.selection.selectors.LocalTokenSelector
+import com.r3.corda.lib.tokens.selection.selectors.Selector
+import com.r3.corda.lib.tokens.selection.services.VaultWatcherService
+import net.corda.core.cordapp.CordappConfig
+import net.corda.core.cordapp.CordappConfigException
+import net.corda.core.node.ServiceHub
+import org.slf4j.LoggerFactory
+
+const val CACHE_SIZE_DEFAULT = 1024 // TODO Return good default, for now it's not wired, it will be done in separate PR.
```
roger3cev: 👍

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/internal/selection/DatabaseTokenSelection.kt
```diff
@@ -111,7 +114,7 @@ class DatabaseTokenSelection(
         val stateAndRefs = mutableListOf<StateAndRef<FungibleToken>>()
         for (retryCount in 1..maxRetries) {
             val additionalCriteria = if (queryBy?.issuer != null) {
-                tokenAmountWithIssuerCriteria(requiredAmount.token, queryBy.issuer)
+                tokenAmountWithIssuerCriteria(requiredAmount.token, queryBy.issuer!!) //TODO !!
```
roger3cev: What's this bit ?

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/internal/ExternalIdIndexUtils.kt
```diff
@@ -0,0 +1,75 @@
...
+import java.util.*
+
+/** Table names. */
+val publicKeyHashToExternalId = "pk_hash_to_ext_id_map"
+
+/** Column names. */
+val publicKeyHashToExternalId_externalId = "external_id"
+val publicKeyHashToExternalId_publicKeyHash = "public_key_hash"
+
+sealed class Holder {
```
roger3cev: Again let's revisit the design of this before we release.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/internal/ExternalIdIndexUtils.kt
```diff
@@ -0,0 +1,75 @@
...
+                MappedIdentity(uuid)
+            } else {
+                UnmappedIdentity
+            }
+        }
+    }
+}
+
+// TODO after corda 4.3 we will be able to use cache introduced by this PR https://github.com/corda/corda/pull/5357
+fun lookupExternalIdFromKey(owningKey: PublicKey, serviceHub: ServiceHub): Holder {
```
roger3cev: There will be an API on identity service for doing this in 4.3.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/internal/ExternalIdIndexUtils.kt
```diff
@@ -0,0 +1,75 @@
...
+/** Table names. */
+val publicKeyHashToExternalId = "pk_hash_to_ext_id_map"
+
+/** Column names. */
+val publicKeyHashToExternalId_externalId = "external_id"
+val publicKeyHashToExternalId_publicKeyHash = "public_key_hash"
+
+sealed class Holder {
+    data class KeyIdentity(val owningKey: PublicKey) : Holder() // Just public key
+    object UnmappedIdentity : Holder() // For all keys that are unmapped
```
roger3cev: what's the different between unmapped identity and key identity?

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/services/VaultWatcherService.kt
```diff
@@ -30,35 +32,43 @@ val UNLOCKER: ScheduledExecutorService = Executors.newSingleThreadScheduledExecu
...
 
 @CordaService
-class VaultWatcherService(tokenObserver: TokenObserver? = null) : SingletonSerializeAsToken() {
-
+class VaultWatcherService(private val tokenObserver: TokenObserver, private val serviceHub: ServiceHub) : SingletonSerializeAsToken() {
     enum class IndexingType {
-        EXTERNAL_ID, PUBLIC_KEY
+        EXTERNAL_ID, // external id (all keys registered to this id as an owner) + token class + token identifier
+        PUBLIC_KEY, // Public key + token class + token identifier
+        TOKEN // Just token class + token identifier without owner
```
roger3cev: TOKEN_ONLY ?

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/services/VaultWatcherService.kt
```diff
@@ -30,35 +32,43 @@ val UNLOCKER: ScheduledExecutorService = Executors.newSingleThreadScheduledExecu
...
-                }
-            } else {
-                { _, _ ->
-                    throw IllegalStateException("Only IndexingType.PUBLIC_KEY available on Corda V4")
-                }
+            val configOptions = if (providedConfig == null) {
+                InMemorySelectionConfig.parse(config)
+            } else providedConfig
+
+            val ownerProvider: ((StateAndRef<FungibleToken>, ServiceHub) -> Holder) = when (configOptions.indexingStrategy) {
```
roger3cev: This is a bit "special" :) Perhaps we can tidy up at some point. Not essential for now though.

roger3cev:
I'll spend a bit more time looking through this code in my IDE and think about some of the design stuff. Cheers!

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/selectors/Selector.kt
kasiastreich: We can, it will make things easier for now.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/selectors/Selector.kt
kasiastreich: I didn't, but also I didn't want that module to depend on workflows, so I had to deconstruct that type. Ideally it should be token independent in the future.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/config/InMemorySelectionConfig.kt
kasiastreich: I would argue that we don't need allowShortfall, I don't see sensible use case for it.
Also we have sensible default already.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/config/InMemorySelectionConfig.kt
kasiastreich: See next commit where exactly that happens.

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/internal/selection/DatabaseTokenSelection.kt
kasiastreich: ah, forgot to cleanup after refactor, good spot

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/internal/ExternalIdIndexUtils.kt
kasiastreich: it's internal for indexing

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/internal/ExternalIdIndexUtils.kt
kasiastreich: yeah, as the comment on that function says

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/internal/ExternalIdIndexUtils.kt
kasiastreich: key identity is used for indexing by public key
unmapped identity is everything that doesn't map to external id - used for external identity indexing

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/services/VaultWatcherService.kt
kasiastreich: ok

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/services/VaultWatcherService.kt
kasiastreich: What is special about it? This is function on service construction, you can pass indexing only via config options (services have access only to `AppServiceHub`). To be able to index by anything you would have to have some translation layer between config and owner provider.

modules/selection/build.gradle
```diff
@@ -35,6 +35,10 @@ dependencies {
     cordaCompile ("$corda_release_group:corda-core:$corda_release_version"){
         changing = true
     }
+    // Corda dependencies.
+    cordaCompile ("$corda_release_group:corda-node-api:$corda_release_version"){
+        changing = true
```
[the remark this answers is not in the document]
kasiastreich: `net.corda.nodeapi.internal.persistence.CordaPersistence` is needed because bug in Corda that doesn't expose database on vault observable thread

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/TokenQueryBy.kt
```diff
@@ -1,14 +1,12 @@
...
+import com.r3.corda.lib.tokens.selection.internal.Holder
 import net.corda.core.contracts.StateAndRef
-import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.Party
 import net.corda.core.node.services.vault.QueryCriteria
-import java.util.*
 
-// TODO I don't like Any in owner to be honest, it should be sealed class for either external id or public key/Abstract party
-data class TokenQueryBy(val holder: Any? = null, val issuer: Party? = null, val predicate: (StateAndRef<FungibleToken>) -> Boolean = { true }, val queryCriteria: QueryCriteria? = null) {
+data class TokenQueryBy(val holder: Holder? = null, val issuer: Party? = null, val predicate: (StateAndRef<FungibleToken>) -> Boolean = { true }, val queryCriteria: QueryCriteria? = null) {
```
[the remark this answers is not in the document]
kasiastreich: yes, agree, it is here because of PoC and because we don't want to break API for token-sdk flows that take queryCriteria. In the future ideally we won't have `queryCriteria` at all in flows and query only on issuer, token type, etc. Under the hood it  should get translated to either database or in memory token selection.

PR #155, opened 2019-11-15
NOTICK: Ensure deterministic modules are compiled for Java 8.

chrisr3:
Small Gradle enhancement to ensure deterministic modules are still compiled for Java 8, even if/when you eventually port to Java 11.

Note that Kotlin 1.2.x can _only_ be compiled to either Java 6 or Java 8 byte-code. Compiling for Java 9+ requires Kotlin >= 1.3.30.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,5 @@
+import static org.gradle.api.JavaVersion.VERSION_1_8
+
 buildscript {
     ext {
         corda_release_group = 'net.corda'
@@ -77,7 +79,7 @@ subprojects {
         kotlinOptions {
             languageVersion = "1.2"
             apiVersion = "1.2"
-            jvmTarget = "1.8"
+            jvmTarget = VERSION_1_8
             javaParameters = true   // Useful for reflection.
         }
     }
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -77,7 +79,7 @@ subprojects {
         kotlinOptions {
             languageVersion = "1.2"
             apiVersion = "1.2"
-            jvmTarget = "1.8"
+            jvmTarget = VERSION_1_8
```
roger3cev: This sets the target for all JARs in the repo. Potentially we might target later versions for non-contract JARs, when the JAva 11 work is done?

chrisr3: This _particular_ line just replaces "1.8" with its equivalent `JavaVersion` object, and "yes" this is configuring the Kotlin compiler to generate Java 8 byte-code across all sub-projects.

You would want to change this to `VERSION_11`, but have `deterministic.gradle` override it with `VERSION_1_8` for the `contract` and `money` modules.

roger3cev: ok sounds good - just checking. Cheers!

PR #160, opened 2019-12-03
Add converter for tokenClass schema

kasiastreich:
Add converter for both FungibleTokenSchema and NonFungibleTokenSchema,
that deals with tokenClass: Class<*>


roger3cev:
Approved but I guess we don't have a test for this? :(

PR #173, opened 2020-01-10
Use unlocked and specified softlocking condition instead of unlocked …

roger3cev:
…only, so we can select tokens which were issued in the same flow that the selection is taking place in. This is because states added to the vault are automatically locked to avoid other flows "stealing" them, and causing unexpected behaviour.


roger3cev:
build

PR #174, opened 2020-01-15
Fix for CORDA 3524

roger3cev:
Updated the ERROR to a WARN when an indexing strategy for the in-memory token selection config has not been specified. The message is also now more informative.


roger3cev:
@roastario nudge

PR #178, opened 2020-02-11
Lts doc updates

nimmaj:
Hey all - pls find some proposed changes to the LTS docs in t-sdk 1.1 based on some things we found using them.

Imagine you'll (quite correctly :-) ) take issue with the lateinit var and the horrific map example for tests. The latter i'm open to all suggestions.  The former we did because we found that the LTS was used in deeply nested structures and as such passing it down the call hierarchy wasn't very nice.  Though obviously nor is a var.  Any thoughts?

Thanks again for this bit of work.  Amazing to have a fast, threadsafe token selector.

### DIFF
--- a/docs/InMemoryTokenSelection.md
+++ b/docs/InMemoryTokenSelection.md
@@ -162,3 +162,43 @@ private val tokenSelectionConfig = mapOf<String, Any>("stateSelection" to
                 mapOf<String, Any>("enabled" to true, "cacheSize" to 1024, "indexingStrategies" to listOf("EXTERNAL_ID"))))
 val TokenSelectionCordapps: Set<TestCordapp> =  setOf(TestCordapp.findCordapp("com.r3.corda.lib.tokens.selection")).map{ it.withConfig(tokenSelectionConfig) }.toSet()
 ```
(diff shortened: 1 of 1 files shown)


roastario:
👌 thanks a lot!

docs/InMemoryTokenSelection.md
```diff
@@ -148,4 +150,61 @@ val (inputs, outputs) = localTokenSelector.generateMove(
...
+mimic the existing behaviour of the db token selection you can do something like:
+
+```kotlin
+package io.mycompany.common.workflows.flows
+
+import co.paralleluniverse.fibers.Suspendable
+import com.r3.corda.lib.tokens.selection.memory.selector.LocalTokenSelector
+import net.corda.core.flows.FlowLogic
+
+abstract class SdxTokenReleaseFlow<out T>  : FlowLogic<T>() {
```
kasiastreich: I would rename that class from SDX to something else :)

nimmaj: oops - good spot - i thought i'd done that!  just the package though it would seem!

PR #187, opened 2020-03-11
EG-685: Wrap utilities in class for access from java

rogersanick:


anixon604:
@rogersanick Just doing a quick comparison of calls from the Java side.

Kotlin call:
val n = getPreferredNotary(serviceHub)

Java calls:

        // Original
        Party n = NotaryUtilitiesKt.getPreferredNotary(getServiceHub(), NotaryUtilitiesKt.firstNotary());
        
        // As Proposed
        Party n = new PreferredNotary(getServiceHub()).get();
        
Other possibilities

        // Companion Object with @JvmStatic
        Party n = PreferredNotary2.get(getServiceHub());

        // Overarching single class for single import with @JvmStatic functions
        Party n = TokenUtilities.getPreferredNotary(getServiceHub());

Wondering if the best experience might be something like the last. Where we have an SDK wide Utility class with static methods piping in to the Kot funs? What do you think?

anixon604:
Let's add
- @JvmStatic to getInstance on DigitalCurrency and FiatCurrency
- @JvmOverloads constructor(...) to the class headers of TokenQueryBy and DatabaseTokenSelection

anixon604:
And sync filenames to match Kotlin calls
@file:JvmName("...Utilities")

anixon604:
Found a few more lingering for annotation - for completeness. I think we'll be good for the merge after these.

In Modules:
	• Selector @JvmOverload funs
	• InMemoryTokenSelectionConfig @JvmOverload const, @JvmStatic parse
• LocalTokenSelector @JvmOverload const, @JvmStatic getVaultObs

In Workflows:
	• UpdateEvolveableTokenFlow /@JvmOverloads
	• IssueTokenUtilities @JvmName_
	• MoveTokenUtilities @JvmName_
	• RedeemFlowUtilities @JvmName_
	• EvolveableTokeUtilities @JvmName_
	• DistributionUtilities @JvmName_
	• CheckUtilities @JvmName_
	• FlowUtilities @JvmName_
         • QueryUtilities @JvmName_

anixon604:
Just missing the Selector.kt @JvmOverload funs
selectTokens and generateMove, have defaults on tokenQueryBy and UUID

So something like DatabaseTokenSelection(serviceHub).XXX are still diff.

PR #188, opened 2020-03-13
NOTICK: Make it easier for Java developers to build Tokens correctly.

chrisr3:
A `Token` _must_ know which attachment defines its `TokenType`. Kotlin developers have a default constructor parameter to help them, but Java developers need to call the appropriate Magic Function themselves. Help Java developers out by creating suitable secondary constructors.

### DIFF
--- a/contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/states/FungibleToken.kt
+++ b/contracts/src/main/kotlin/com/r3/corda/lib/tokens/contracts/states/FungibleToken.kt
@@ -35,7 +35,7 @@ import net.corda.core.schemas.QueryableState
  * @property holder the [AbstractParty] which has a claim on the issuer of the [IssuedTokenType].
  */
 @BelongsToContract(FungibleTokenContract::class)
-open class FungibleToken(
+open class FungibleToken @JvmOverloads constructor(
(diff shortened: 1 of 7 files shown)


chrisr3:
> Hey @chrisr3, this looks good to me! Just confirming, have you tried accessing either of these constructors from java code?
> 
> If you look at my [PR](https://github.com/corda/token-sdk/pull/187), I've add sanity check units test for access from Java.

No, I just applied the standard Kotlin annotation for the "default parameters" case. (And if that doesn't work then it's a _Kotlin_ bug.)

BTW, it's better to use the JUnit or AssertJ `assertEquals()`, `assertTrue()`, `assertThat()` etc functions than Kotlin's `assert()` because `assert()` is implemented as:
```kotlin
@kotlin.internal.InlineOnly
public inline fun assert(value: Boolean, lazyMessage: () -> Any) {
    if (_Assertions.ENABLED) {
        if (!value) {
            val message = lazyMessage()
            throw AssertionError(message)
        }
    }
}
```
and so will not test anything unless `_Assertions.ENABLED` is true.

chrisr3:
@rogersanick I've added some simple Java test classes that actually _look_ like Java rather than decompiled Kotlin. Feel free to merge them or not.

PR #192, opened 2020-03-23
EG-710: Create a builder class to replicate Tokens DSL

rogersanick:


workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
```diff
@@ -0,0 +1,171 @@
...
+    fun withAmount(bigDecimalAmount: BigDecimal) = initializeAmountOrThrow { this.bigDecimalAmount = bigDecimalAmount }
+
+    /**
+     * Replicates the Kotlin DSL [of] infix function. After supplying a [TokenType] to the builder
+     * the developer will be able to call [resolveAmountTokenType] to access the [amountTokenType]
+     * member property.
+     *
+     * @param t The token type that will be used to initialize the [amountTokenType] member property.
+     */
+    fun <T: TokenType> of(t: T): TokenBuilder = when (getInitializedAmount()) {
```
ivanr3: The builder should be converting itself to a value only when you call `build() function, not when you supply values to it.

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * Replicates the Kotlin DSL [heldBy] infix function. After supplying a [Party] to the builder
+     * representing the identity of the holder of a new fungible token, the developer will be able
+     * to call [resolveAmountIssuedTokenType] to access the [fungibleToken] member property.
+     *
+     * @param party The identity of the holder that will be used to initialize the [amountIssuedTokenType]
+     * member property.
+     */
+    fun heldBy(party: Party): TokenBuilder {
```
ivanr3: same here, wait until all vlaues are supplied first, then verify

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
```diff
@@ -0,0 +1,171 @@
...
+        }
+        return this
+    }
+
+    /**
+     * A simple get function that resolves to the [amountTokenType] value.
+     * This function will throw a not initialized exception if the appropriate
+     * builder methods have not been called: [withAmount], [of].
+     */
+    fun resolveAmountTokenType() = this.amountTokenType
```
ivanr3: Maybe `buildToken`, `buildFungibleToken`, `buildIssuedToken`?

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
```diff
@@ -0,0 +1,171 @@
...
+     * Replicates the Kotlin DSL [issuedBy] infix function. After supplying a [Party] to the builder
+     * representing the identity of the issuer the developer will be able to call [resolveAmountTokenType]
+     * to access the [amountIssuedTokenType] member property.
+     *
+     * @param party The issuing identity that will be used to initialize the [amountIssuedTokenType] member property.
+     */
+    fun issuedBy(party: Party): TokenBuilder {
+        this.amountIssuedTokenType = when {
+            ::amountTokenType.isInitialized -> { this.amountTokenType.issuedBy(party) }
+            else -> { throw TokenBuilderException("An amountTokenType has not been initialized") }
```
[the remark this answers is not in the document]
rogersanick: That makes sense generally! For `amount` specifically in this context, it's conversion to `Long` depends on the information provided in the `TokenType`. 

I actually refactored to convert to Long automatically (only having one amount member property), but then the conversion was inaccurate and did not behave the same as the Kotlin DSL.

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
rogersanick: Makes sense! Fixed this.

rogersanick:
@ivanr3 - updated with feedback. Let me know what you think.

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
```diff
@@ -0,0 +1,157 @@
...
+        }
+    } catch (ex: UninitializedPropertyAccessException) {
+        throw TokenBuilderException("An amount value has not been provided to the builder.")
+    }
+
+    /**
+     * Builds an [Amount] of an [IssuedTokenType]. This function will throw a [TokenBuilderException]
+     * if the appropriate builder methods have not been called: [withAmountValue], [of], [issuedBy].
+     */
+    fun buildAmountIssuedTokenType(): Amount<IssuedTokenType> = when {
```
ivanr3: Any reason why not to call it just `buildIssuedToken`?

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * A utility class designed for Java developers to more easily access Kotlin DSL
+ * functions to build token types.
+ *
+ * This function vaguely follows a builder pattern, design choices were made
+ * to emulate Kotlin syntax as closely as possible for an easily transferable
+ * developer experience in Java.
+ */
+class TokenBuilder {
+    private var longAmount: Long? = null
```
ivanr3: Maybe lets use BigDecimal as the lowest common dinaminator for all other data types? Then there won't be a need for Lond, double, etc.

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
rogersanick: I think `buildIssuedToken` could be confusing because what's returned isn't actually a token yet. I agree that `buildAmountIssuedTokenType` is a little verbose, but I think in this case it's appropriate. Happy to change if you think!

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
rogersanick: Very interesting, I'll try this one out.

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * A utility class designed for Java developers to more easily access Kotlin DSL
+ * functions to build token types.
+ *
+ * This function vaguely follows a builder pattern, design choices were made
+ * to emulate Kotlin syntax as closely as possible for an easily transferable
+ * developer experience in Java.
+ */
+class TokenBuilder {
+    private var bigDecimalAmount: BigDecimal? = null
```
ivanr3: it can be just `amount` now I guess :)

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
```diff
@@ -0,0 +1,113 @@
...
+    private lateinit var tokenType: TokenType
+    private lateinit var issuer: Party
+    private lateinit var holder: Party
+
+    /**
+     * Set the [bigDecimalAmount] member property of the builder using a provided [Long].
+     *
+     * @param longAmount The [Long] that will be converted to set the [bigDecimalAmount] member property
+     */
+    fun withAmountValue(longAmount: Long) = this.apply { bigDecimalAmount = BigDecimal.valueOf(longAmount) }
```
ivanr3: `withAmount` might be simpler

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Set the [bigDecimalAmount] member property of the builder using a provided [Int].
+     *
+     * @param intAmount The [Int] that will be converted to set the [bigDecimalAmount] member property
+     */
+    fun withAmountValue(intAmount: Int) = this.apply { bigDecimalAmount = BigDecimal.valueOf(intAmount.toLong()) }
```
ivanr3: You don't have to explicitly convert `toLong` here - BigDecimal already has overriden constructors for doubles, ints and longs

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
ivanr3: What about `buildFungibleToken` then? How different is it from others?

workflows/src/test/java/com/r3/corda/lib/tokens/workflows/TokenBuilderTests.java
```diff
@@ -0,0 +1,132 @@
...
+                    .buildAmountIssuedTokenType();
+            assert(false);
+        } catch(Exception ex) {
+            assert(ex instanceof TokenBuilderException);
+            assert(ex.getLocalizedMessage().equals("An token issuer has not been provided to the builder."));
+        }
+    }
+
+    @Test
+    public void UnableToRetrieveFungibleTokenWithoutHolder() throws Exception {
```
ivanr3: It's better to be more specific about types of exceptions the method throws. For example here would be better to use `throws TokenBuilderException`

workflows/src/test/java/com/r3/corda/lib/tokens/workflows/TokenBuilderTests.java
```diff
@@ -0,0 +1,132 @@
...
+        StartedMockNode alice = node(aliceX500Name);
+        Party aliceParty = alice.getInfo().getLegalIdentities().get(0);
+        try {
+            new TokenBuilder()
+                    .withAmountValue(new Long(1))
+                    .of(FiatCurrency.getInstance("USD"))
+                    .issuedBy(aliceParty)
+                    .buildFungibleToken();
+            assert(false);
+        } catch(Exception ex) {
```
ivanr3: why not to catch `TokenBuilderException`?

workflows/src/test/java/com/r3/corda/lib/tokens/workflows/TokenBuilderTests.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Test
+    public void UnableToRetrieveAmountIssuedTokenTypeWithoutIssuer() throws Exception {
+        try {
+            new TokenBuilder()
+                    .withAmountValue(new Long(1))
+                    .of(FiatCurrency.getInstance("USD"))
+                    .buildAmountIssuedTokenType();
+            assert(false);
+        } catch(Exception ex) {
```
ivanr3: same here

workflows/src/test/java/com/r3/corda/lib/tokens/workflows/TokenBuilderTests.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Test
+    public void UnableToRetrieveAmountTokenTypeWithoutTokenType() throws Exception {
+        try {
+            new TokenBuilder()
+                    .withAmountValue(new Long(1))
+                    .buildAmountTokenType();
+            assert(false);
+        } catch(Exception ex) {
```
ivanr3: same here

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
rogersanick: This is the only method that returns a state - specifically a `FungibleToken` which extends `FungibleState<IssuedTokenType>` while buildAmountIssuedTokenType and buildAmountTokenType both just return `Amounts`

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
rogersanick: Good point!

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
rogersanick: I thought this might be confusing given that `Amount` is already a class. In retrospect I think `withAmount` is still more clear. I'll revert to withAmount for now.

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
rogersanick: Fixed!

rogersanick: Had to use a different invocation for Ints.

workflows/src/test/java/com/r3/corda/lib/tokens/workflows/TokenBuilderTests.java
rogersanick: Fixed! I hadn't added the @Throws(TokenBuilderException::class) to the Kotlin function which was causing a compilation error.

workflows/src/test/java/com/r3/corda/lib/tokens/workflows/TokenBuilderTests.java
rogersanick: Fixed.

workflows/src/test/java/com/r3/corda/lib/tokens/workflows/TokenBuilderTests.java
rogersanick: Fixed.

workflows/src/test/java/com/r3/corda/lib/tokens/workflows/TokenBuilderTests.java
rogersanick: Fixed.

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
```diff
@@ -0,0 +1,115 @@
...
+     * @param doubleAmount The [Double] that will be converted to set the [amount] member property
+     */
+    fun withAmount(doubleAmount: Double) = this.apply { amount = BigDecimal.valueOf(doubleAmount) }
+
+    /**
+     * Set the [amount] member property of the builder using a provided [BigDecimal].
+     *
+     * @param bigDecimal The [BigDecimal] that will be used to set the [amount] member property
+     */
+    fun withAmount(bigDecimal: BigDecimal) = this.apply { amount = bigDecimal }
```
ivanr3: bigDecimalAmount to be consistent

workflows/src/test/kotlin/com/r3/corda/lib/tokens/workflows/DiamondWithTokenScenarioTests.kt
```diff
@@ -1,6 +1,7 @@
 package com.r3.corda.lib.tokens.workflows
 
 import com.r3.corda.lib.tokens.contracts.states.NonFungibleToken
+import com.r3.corda.lib.tokens.money.USD
```
ivanr3: this import is not needed I believe

rogersanick: Should I remove this in a seperate PR? It's related to DiamondWithTokenScenarioTests?

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/TokenBuilder.kt
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Builds an [Amount] of a [TokenType]. This function will throw a [TokenBuilderException]
+     * exception if the appropriate builder methods have not been called: [withAmount], [of].
+     */
+    @Throws(TokenBuilderException::class)
+    fun buildAmountTokenType(): Amount<TokenType> = when {
+        !::tokenType.isInitialized -> { throw TokenBuilderException("A Token Type has not been provided to the builder.") }
+        amount != null -> { amount!! of tokenType }
+        else -> { throw TokenBuilderException("An amount value has not been provided to the builder.") }
```
ivanr3: I'd reorder verification statements to do all assertions first and then return the value in the else block, i.e.
```
when {
        !::tokenType.isInitialized -> throw TokenBuilderException("A Token Type has not been provided to the builder.")
        amount == null -> throw TokenBuilderException("An amount value has not been provided to the builder.") 
        else -> {  amount!! of tokenType  }
```

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/NonFungibleTokenBuilder.kt
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * A utility class designed for Java developers to more easily access Kotlin DSL
+ * functions to build non-fungible tokens and their component classes.
+ *
+ * This function vaguely follows a builder pattern, design choices were made
+ * to emulate Kotlin syntax as closely as possible for an easily transferable
+ * developer experience in Java.
+ */
+class NonFungibleTokenBuilder {
```
ivanr3: ny reason why we can't use the same builder for both fungible and non-fungible tokens? Both of the builders seem to have the same fields...

workflows/src/test/java/com/r3/corda/lib/tokens/workflows/NonFungibleTokenBuilderTests.java
```diff
@@ -0,0 +1,93 @@
...
+import org.junit.Test;
+
+import java.security.KeyPairGenerator;
+import java.security.NoSuchAlgorithmException;
+import java.security.PublicKey;
+
+public class NonFungibleTokenBuilderTests {
+
+    @Test
+    public void NonFungibleTokenBuilderResolvesWithoutThrowing() throws TokenBuilderException, NoSuchAlgorithmException {
```
ivanr3: method names should be starting with a lowercase.

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/utilities/NonFungibleTokenBuilder.kt
anixon604: This was a usability compromise. Requested this so that it's more intuitive for novice users to quickly discern what elements they need to add to build fungible vs non-fungible.

ivanr3: ok makes sense

workflows/src/test/java/com/r3/corda/lib/tokens/workflows/NonFungibleTokenBuilderTests.java
rogersanick: Adding fix then will merge!

PR #194, opened 2020-03-25
make Fiat and Digital Currencies not constructable

roastario:


roastario:
temporary step before removing money entirely.

PR #199, opened 2020-05-05
Make redeem flow handlers open

adagys:
…to allow adding custom processing logic for redeemed tokens on the issuer`s side.

Right now there is no way of tracking or processing redeemed tokens for the issuer without using the deprecated internal transaction RPC feed or copying and re-implementing the redeem flows.

The simplest fix seems to be making the handlers open for subclassing.


workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/flows/redeem/ConfidentialRedeemFungibleTokensFlowHandler.kt
```diff
@@ -1,15 +0,0 @@
-package com.r3.corda.lib.tokens.workflows.flows.redeem
-
-import co.paralleluniverse.fibers.Suspendable
-import net.corda.core.flows.FlowLogic
-import net.corda.core.flows.FlowSession
-
-/**
- * Responder flow to [ConfidentialRedeemFungibleTokensFlow].
- */
-class ConfidentialRedeemFungibleTokensFlowHandler(val otherSession: FlowSession) : FlowLogic<Unit>() {
```
ivanr3, line 10: Hi @adagys. Thanks for the PR . What's the reason for removing this class?

adagys: It seems to have lost its purpose, currently it just runs a subflow. I made the `ConfidentialRedeemFungibleTokensHandler` call `RedeemTokensFlowHandler` directly.

ivanr3:
It looks good to me. @roastario could you please have a look whether it rings any bells?

workflows/src/main/kotlin/com/r3/corda/lib/tokens/workflows/flows/redeem/ConfidentialRedeemFungibleTokensFlowHandler.kt
roastario: @adagys - the reason we leave these flows in place is due to the nature of this cordapp it's an SDK and so people will usually invoke these handlers as part of their own responder, could we put it back?

adagys: @roastario All the other `TokenFlowHandler`s just call `RedeemTokensFlowHandler` directly, but here we have `ConfidentialRedeemFungibleTokensHandler -> ConfidentialRedeemFungibleTokensFlowHandler -> RedeemTokensFlowHandler`

People could inline `ConfidentialRedeemFungibleTokensHandler` instead. But if this is considered API breakage I'm happy to put it back.

PR #203, opened 2020-05-10
Fixed money not being fatjarred

ivanr3:


roastario:
how extreeeeemely odd.

PR #206, opened 2020-06-06
Put Workflows Jar on a diet - exclude all corda runtime files

roastario:


workflows/build.gradle
```diff
@@ -2,6 +2,11 @@ apply plugin: 'kotlin-jpa'
 apply plugin: 'net.corda.plugins.quasar-utils'
 apply plugin: 'net.corda.plugins.cordapp'
 
+
```
chrisr3, line 5: This entire file ... _WTF?!_
_Why?_

If there isn't an easier way of doing this then _Gradle Must DIE!!!_

roastario: It's an edge case in our fatjarring. 

We want to compile in the selection project, but the selection project `cordaCompile` Corda components, which when you don't use the jar approach results in the entire Corda runtime being included in the fatJar due to transitive dependencies. 

The next step is to just move all the logic into workflows.. which I'm seriously considering.

ivanr3: Yeah, probably would make sense to do so, given that money and selection aren't separate cordapps anymore and unlikely to be used outside of the token sdk.

PR #212, opened 2020-09-03
Add serialization env to OwnerMigration

blsemo:
This is an idea how a serialization env could be added to make the migration work in a tool environment. It's not clear that this is the right, nor have I tested that this particular environment would make it actually work.
It's also quite a hack as it uses Corda internal classes/functions that are not guaranteed to be stable and are not part of Corda's API.


roastario (approved):
LGTM

PR #215, opened 2020-10-13
CID-1400 added InsufficientNotLockedBalanceException

ischasny:
* Added InsufficientNotLockedBalanceException that is thrown when there is not enough not locked tokens is available. 
* Updated both in memory any DB selectors.
* Added unit tests


roastario (approved):
Important caveat is that for the InMemory one, due to it being concurrent - it's possible that the Exception thrown might be "out of date" due to tokens being added during selection - but a proper edge case.

PR #216, opened 2020-10-13
CID-1400 throw InsufficientNotLockedBalanceException when there is not enough not locked tokens available

roastario:
…


ivanr3 (approved):
LGTM

PR #241, opened 2023-03-26
ENT-9608: Wait for service lifecycle event before start token loading and handle result set re-sync

adelel1:
Changes made:

Now wait for either ServiceLifecycleEvent.BEFORE_STATE_MACHINE_START or ServiceLifecycleEvent.STATE_MACHINE_STARTED before we start tokens loading into memory. By waiting for the above events token loading takes place after the vault is initialised.

When reading vault states into query pages if we detect that a page has become out of sync, then we reread the last page and determine if we can sync with that page, and so on until we find a page we can sync with. We establish sync points by slightly overlapping the database query.

Regarding the 2 ignored test methods. The VaultWatcherService now calls the register method of AppServiceHub. The test uses a AppServiceHub mock defined in Corda, unfortunately in the mock method this throws a not implemented exception. Which is the exception we get when test method is run. Couldn't define my own mock as the mock implementation adds the VaultWatcherService to Corda service list. One solution would be to update Corda to not throw exceptions when not implemented (by the mock). 

Had to revert having having the vault changes run on the UPDATER thread. Tests started to fail. Problem is a new token is created, this then causes the event to be queued to run in the UPDATER thread. But the next line of code in the test selects from the VaultWatcherService so is assuming the VaultWatcherService has been updated (which it may not have been, as the update is occurring in a different thread now), so causing failure.

### DIFF
--- a/modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
+++ b/modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
@@ -249,7 +249,7 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
         tokenObserver.source.doOnError {
             LOG.error("received error from observable", it)
         }
-        tokenObserver.source.subscribe{UPDATER.submit{ onVaultUpdate(it)}}
+        tokenObserver.source.subscribe(::onVaultUpdate)
     }
 
     private fun processToken(token: StateAndRef<FungibleToken>, indexingType: IndexingType): TokenIndex {
--- a/modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
+++ b/modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
@@ -15,6 +15,7 @@ import net.corda.core.contracts.StateAndRef
 import net.corda.core.internal.uncheckedCast
 import net.corda.core.node.AppServiceHub
 import net.corda.core.node.services.CordaService
+import net.corda.core.node.services.ServiceLifecycleEvent
 import net.corda.core.node.services.Vault
 import net.corda.core.node.services.vault.DEFAULT_PAGE_NUM
 import net.corda.core.node.services.vault.PageSpecification
@@ -23,11 +24,13 @@ import net.corda.core.serialization.SingletonSerializeAsToken
 import net.corda.core.utilities.contextLogger
 import rx.Observable
 import java.time.Duration
+import java.util.ArrayDeque
 import java.util.concurrent.*
 import java.util.concurrent.atomic.AtomicReference
 import java.util.concurrent.locks.ReentrantReadWriteLock
 import kotlin.concurrent.read
 import kotlin.concurrent.write
+import kotlin.math.floor
 
 val UPDATER: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
 val EMPTY_BUCKET = TokenBucket()
@@ -67,7 +70,18 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
 
     }
 
-    constructor(appServiceHub: AppServiceHub) : this(getObservableFromAppServiceHub(appServiceHub), InMemorySelectionConfig.parse(appServiceHub.getAppContext().config))
+    var tokenLoadingStarted = false
+    constructor(appServiceHub: AppServiceHub) : this(getObservableFromAppServiceHub(appServiceHub), InMemorySelectionConfig.parse(appServiceHub.getAppContext().config)) {
+        appServiceHub.register(AppServiceHub.SERVICE_PRIORITY_NORMAL) { event ->
+            when (event.name) {
+                ServiceLifecycleEvent.BEFORE_STATE_MACHINE_START.toString(),
+                ServiceLifecycleEvent.STATE_MACHINE_STARTED.toString() -> if (!tokenLoadingStarted) {
+                    tokenLoadingStarted = true
+                    tokenObserver.startLoading(::onVaultUpdate)
+                }
+            }
+        }
+    }
 
     companion object {
         val LOG = contextLogger()
@@ -95,10 +109,9 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
 
 
             val pageSize = 1000
-            var currentPage = DEFAULT_PAGE_NUM
             val (_, vaultObservable) = appServiceHub.vaultService.trackBy(
                     contractStateType = FungibleToken::class.java,
-                    paging = PageSpecification(pageNumber = currentPage, pageSize = pageSize),
+                    paging = PageSpecification(pageNumber = DEFAULT_PAGE_NUM, pageSize = pageSize),
                     criteria = QueryCriteria.VaultQueryCriteria(status = Vault.StateStatus.ALL),
                     sorting = sortByStateRefAscending())
 
@@ -111,18 +124,15 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
                     UPDATER.submit {
                         try {
                             var shouldLoop = true
+                            var startResultSetIndex = 0
                             while (shouldLoop) {
-                                val newlyLoadedStates = appServiceHub.vaultService.queryBy(
-                                        contractStateType = FungibleToken::class.java,
-                                        paging = PageSpecification(pageNumber = currentPage, pageSize = pageSize),
-                                        criteria = QueryCriteria.VaultQueryCriteria(),
-                                        sorting = sortByStateRefAscending()
-                                ).states.toSet()
+                                val queryResult = queryVaultForStates(appServiceHub, startResultSetIndex)
+                                val newlyLoadedStates = queryResult.first.toSet()
+                                startResultSetIndex = queryResult.second
+                                shouldLoop = queryResult.third
                                 LOG.info("publishing ${newlyLoadedStates.size} to async state loading callback")
                                 callback(Vault.Update(emptySet(), newlyLoadedStates))
-                                shouldLoop = newlyLoadedStates.isNotEmpty()
                                 LOG.debug("shouldLoop=${shouldLoop}")
-                                currentPage++
                             }
                             LOG.info("finished token loading")
                         } catch (t: Throwable) {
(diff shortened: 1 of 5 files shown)


modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
```diff
@@ -111,36 +124,131 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
...
+                            syncPoints.addFirst(SyncPoint(states.last(), specification, toResultSetIndex(states.lastIndex, specification)))
+                            val pos = states.indexOf(it.stateAndRef)
+                            it.lastKnownResultSetIndex = toResultSetIndex(pos, specification)
+                            return pos + 1 // Return first unread entry index
+                        }
+                    }
+                    LOG.info("Token loading has become out of sync, will re-sync")
+                    return -1
+                }
+                private fun getNewPageSpecification(): PageSpecification {
```
rick-r3: Is a better name to imply what it does something like `popPriorPageSpecification` ?

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
```diff
@@ -111,36 +124,131 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
...
+                    } else {
+                        pageSpecifications.last
+                    }
+                }
+
+                private fun toResultSetIndex(listEntryIndex: Int, specification: PageSpecification): Int {
+                    return specification.pageSize * (specification.pageNumber-1) + listEntryIndex
+                }
+
+                private fun toListEntryIndex(resultSetIndex: Int, specification: PageSpecification): Int {
```
rick-r3: `toIndexWithinPage` ?

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
```diff
@@ -111,36 +124,131 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
...
+                    if (syncPoints.isEmpty()) {
+                        syncPoints.addFirst(SyncPoint(states.last(), specification, toResultSetIndex(states.lastIndex, specification)))
+                        return 0
+                    }
+                    val highResultSetIndex = toResultSetIndex(states.lastIndex, specification)
+                    val lowResultSetIndex = toResultSetIndex(0, specification)
+
+                    syncPoints.forEach {
+                        val listIndex = toListEntryIndex(it.lastKnownResultSetIndex, specification)
+                        if (listIndex in lowResultSetIndex..highResultSetIndex && states[listIndex] == it.stateAndRef) {
```
rick-r3: This comparison of `listIndex` (index within page) to result set indexes doesn't make any sense.  Surely it doesn't work.   I'd be tempted to remove this `if` and just always fall back to the `states.contains(it.stateAndRef)` (or however it ends up written) as less conditions to write tests for.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
```diff
@@ -186,7 +294,7 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
             for (stateAndRef in stateAndRefs) {
                 val existingMark = __backingMap.putIfAbsent(stateAndRef, PLACE_HOLDER)
                 existingMark?.let {
-                    LOG.warn("Attempted to overwrite existing token ${stateAndRef.ref}, this suggests incorrect vault behaviours")
+                    LOG.warn("Attempted to overwrite existing token ${stateAndRef.ref}, this suggests a result set re-sync occurred")
```
rick-r3: We should reconsider this warning.  Does something break when this condition occurs?  Should the function return or re-do the indexing (as it does now)?   Should it really be a `debug`?

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
```diff
@@ -111,36 +124,131 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
...
+                    return specification.pageSize * (specification.pageNumber-1) + listEntryIndex
+                }
+
+                private fun toListEntryIndex(resultSetIndex: Int, specification: PageSpecification): Int {
+                    return resultSetIndex - ((specification.pageNumber-1) * specification.pageSize)
+                }
             }
             return TokenObserver(emptyList(), uncheckedCast(vaultObservable), ownerProvider, asyncLoader)
         }
+        data class SyncPoint(val stateAndRef: StateAndRef<FungibleToken>, val specification: PageSpecification, var lastKnownResultSetIndex: Int)
```
rick-r3: Maybe `lastKnownResultSetIndex` should be `resultSetIndexOfStateAndRef` ?

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
```diff
@@ -111,36 +124,131 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
...
+                    return if (pageSpecifications.isEmpty()) {
+                        syncPoints.clear()
+                        pageSpecifications.add(initialSpecification)
+                        initialSpecification
+                    } else {
+                        pageSpecifications.last
+                    }
+                }
+
+                private fun toResultSetIndex(listEntryIndex: Int, specification: PageSpecification): Int {
```
rick-r3: I think we should be using `Long` for the return value here.  i.e. in theory it is `Int * Int` so could exceed the size of `Int`.   Then need to ripple through other data structures.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
rick-r3: Should also be a `Long`

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
adelel1: Yes, changed.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
adelel1: Changed.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
adelel1: Oops. Now fixed by just checking the index is in range. I would like to keep this if in. As hopefully it would be the normal case. i.e. when nothing has changed in the list we don't do any extra searching etc. So performance wise should be the same as now.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
adelel1: I've changed to a debug. It is just saying that a duplicate entry has been inserted. This would be because of a resync. Although in the code i do try to minimise the duplicate inserts by moving to the next unread entry (in all cases except one, where we have seen the whole list).

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
adelel1: Changed to a long. Have changed to lastKnownResultSetIndexOfStateAndRef
I do want to keep the lastKnown in the name because if the result set has changed, then this value would be inaccurate, but would be the 'last known' value.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
adelel1: Changed.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
```diff
@@ -111,36 +125,165 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
...
+
+                private fun toResultSetIndex(listEntryIndex: Int, specification: PageSpecification): Long {
+                    return specification.pageSize * (specification.pageNumber-1) + listEntryIndex.toLong()
+                }
+
+                private fun toIndexWithinPage(resultSetIndex: Long, specification: PageSpecification): Int {
+                    return (resultSetIndex - ((specification.pageNumber-1) * specification.pageSize)).toInt()
+                }
+
+                fun seachForLargestSeenStateRef(values: List<StateAndRef<FungibleToken>>, element: StateAndRef<FungibleToken>): Int {
```
rick-r3: Use `Collections.binarySearch()` ?  Otherwise we need to unit test this to death.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
```diff
@@ -111,36 +125,165 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
...
+                                result = midPoint
+                                left = midPoint +1
+                            }
+                            else -> right = midPoint -1
+                        }
+                    }
+                    return result
+                }
+                private fun compare(first: StateAndRef<FungibleToken>, second: StateAndRef<FungibleToken>): Int {
+                    if (first.ref.txhash.bytes.toHexString() < second.ref.txhash.bytes.toHexString()) {
```
rick-r3: `SecureHash` is `Comparable`.  So you can use Kotlin `compareValues(first.ref.txhash, second.ref.txhash)`

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
```diff
@@ -111,36 +125,165 @@ class VaultWatcherService(private val tokenObserver: TokenObserver,
...
+                        return -1
+                    }
+                    if (first.ref.txhash.bytes.toHexString() > second.ref.txhash.bytes.toHexString()) {
+                        return 1
+                    }
+                    if (first.ref.index < second.ref.index) {
```
rick-r3: And then if the secure hashes are equal (compareValues == 0), use `compareValues(first.ref.index, second.ref.index)`

rick-r3 (commented):
Looks thematically correct.  Think you can use more built in stuff for the binary search and comparison.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
adelel1: Done.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
adelel1: Done. Using compareBy now.

modules/selection/src/main/kotlin/com.r3.corda.lib.tokens.selection/memory/services/VaultWatcherService.kt
adelel1: Done. Using compareBy now.

rick-r3 (approved):

PR #246, opened 2023-07-26
ENT-6914 Update to Kotlin 1.8, JDK 17 and Gradle 7.6

arshadm:
This PR updates the token-sdk for the 4.12 release of Corda. This release targets Kotlin 1.8 and JDK 17.

Major changes:
1. Upgrade gradle to 7.6
2. Upgrade Kotlin to 1.8
3. Support for building with JDK 17


.ci/Dockerfile
adelel1, line 1: Why was the Dockerfile deleted?

workflows-integration-test/src/integrationTest/kotlin/com/r3/corda/lib/tokens/ci/DriverBasedTest.kt
adelel1, line 1: Wondering why the CI tests have been moved here?

build.gradle
```diff
@@ -91,25 +98,49 @@ subprojects {
...
 
-    apply plugin: 'kotlin'
+    apply plugin: 'org.jetbrains.kotlin.jvm'
 
     tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile) {
         kotlinOptions {
-            languageVersion = "1.2"
-            apiVersion = "1.2"
-            jvmTarget = VERSION_1_8
+            languageVersion = "1.8"
```
shamsasari, line 132: PR title says 1.9

.ci/Dockerfile
arshadm: Previously the Jenkinsfile used that docker image (JDK8) to build the SDK. We no longer do that, the node has the correct version of Java installed. So no longer needed.

workflows-integration-test/src/integrationTest/kotlin/com/r3/corda/lib/tokens/ci/DriverBasedTest.kt
arshadm: Circular dependency issue. This project has a dependency on confidential-identities, these tests wee part of that project. But the tests need tokens-sdk. This mean the two could not be modified in tandem because of this dependency.

Moving the tests here means that confidential-identities no longer has a dependency on the token-sdk

build.gradle
arshadm: Fixed, changed title.

adelel1 (approved):

PR #249, opened 2023-08-02
ES-770: Update Jenkinsfile to add cron to build project regularly

ronanbrowne:
Add a cron job to this Jenkins file.

Build this project twice a week, Monday and Thursday, at a random minute (H) in the hour of midnight to distribute load.

While this project is supported, there is not a lot of activity here, This change is to avoid situations where this project is not built for weeks/ months and build issues are only discovered on a release day.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,7 +16,7 @@ pipeline {
     options { timestamps() }
 
     triggers {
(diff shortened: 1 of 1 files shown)


yarivus (approved):
LGTM

PR #250, opened 2023-08-08
ENT-10472: Added pom information required for maven publish.

adelel1:
ENT-10472: Added pom information required for maven publish.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -128,7 +128,7 @@ def publishProjects = [
         project(":workflows")
 ]
 
-configure(publishProjects) { subproject ->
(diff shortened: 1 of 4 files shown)


wzur-r3 (approved):
LGTM

wzur-r3 (changes requested):
actually for some reason, the only bit which is not present in the module pom file is the `description` which there already

For some reason `subproject.description` returns `null` in the configure(publishProjects) block

wzur-r3 (approved):
fine to merge and work on better solution in another PR

PR #253, opened 2023-11-06
[ENT-6914] Test case fix for JDK17 upgrade

balwantkothari:
Test case fix for JDK17 upgrade


freighter-tests/src/freighterTest/kotlin/freighter/testing/NullHolderOnObserverTest.kt
```diff
@@ -42,32 +43,32 @@ class NullHolderOnObserverTest : DockerRemoteMachineBasedTest() {
 
 	val freighterHelperCordapp = NodeBuilder.DeployedCordapp.fromClassPath("freighter-cordapp-flows")
 
-	@Test(timeout = 300_000)
+	@Test @Timeout(300_000)
```
shamsasari, line 46: Why do these tests need this new annotation, whilst other tests don't? If we don't need this annotation then  revert.

balwantkothari: There was compilation failures and probably because from library which is being used has been changed

PR #258, opened 2024-03-06
ENT-11502: Updated min platform version and fixed kotlin language version

adelel1:
ENT-11502: Updated min platform version and fixed kotlin language version.


ronanbrowne (approved):
LGTM

PR #266, opened 2024-05-13
ENT-11709,ENT-10933,ENT-11694,ENT-11693,ENT-11531,ENT-10971,ENT-10970,ENT-10972,ENT-10973,ENT-10969,ENT-11482,ENT-10924 - Security updates

chriscochrane:
Updates to address security vulnerabilities reported in the Corda 4.12 release pack.
Many issues are resolved merely by building against the latest Corda Ent 4.12 (snapshot in this PR, which will become 4.12-RC02 in due course).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ import static org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_1_9
 buildscript {
     ext {
         corda_release_group = 'net.corda'
-        corda_release_version = '4.12-RC01'
+        corda_release_version = '4.12-SNAPSHOT'
         confidential_id_release_group = "com.r3.corda.lib.ci"
         confidential_id_release_version = "1.2-RC01"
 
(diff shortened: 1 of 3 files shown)


ronanbrowne (approved):
LGTM, but I thought there was a decision previously always to have satellite projects pointed at the latest Corda RC or release.

Also there is a RC02 due imminently AFAIK so should we just wait for that?

anyway maybe this has already been discussed within C4 team so I don't want to block this.

chriscochrane:
> LGTM, but I thought there was a decision previously always to have satellite projects pointed at the latest Corda RC or release.
> 
> Also there is a RC02 due imminently AFAIK so should we just wait for that?
> 
> anyway maybe this has already been discussed within C4 team so I don't want to block this.

Yes you're right, TBH I jumped the gun a bit with this PR. Just checked with Adel, 4.12-RC02 should be along this week so I'll keep this PR open and then when RC02 is available I'll make the update here.
Note that this needs to be done (and a tokens-sdk RC needs to be created) before any release pack for 4.12-RC02 can be built.

ronanbrowne:
> > LGTM, but I thought there was a decision previously always to have satellite projects pointed at the latest Corda RC or release.
> > Also there is a RC02 due imminently AFAIK so should we just wait for that?
> > anyway maybe this has already been discussed within C4 team so I don't want to block this.
> 
> Yes you're right, TBH I jumped the gun a bit with this PR. Just checked with Adel, 4.12-RC02 should be along this week so I'll keep this PR open and then when RC02 is available I'll make the update here. Note that this needs to be done (and a tokens-sdk RC needs to be created) before any release pack for 4.12-RC02 can be built.

Thanks Chris , ok sounds good

ronanbrowne (approved):

adelel1 (approved):

PR #271, opened 2024-07-08
ES-2505: do not hardcode Artifactory credentials in the source code

wzur-r3:
* removed direct use of a cache repository for Gradle via Artifactory, it is no longer needed


ronanbrowne (approved):
LGTM

PR #274, opened 2024-12-10
ENT-12428,ENT-12430 - Dependency update

chriscochrane:
Updated to build against Corda 4.12.4, which fixes a security vulnerability.


ronanbrowne (approved):
LGTM

